PR #4, opened 2020-05-08
Feature: inter-process propagation

kezhenxu94:
image

### DIFF
--- a/skywalking/trace/__init__.py
+++ b/skywalking/trace/__init__.py
@@ -15,8 +15,6 @@
 # limitations under the License.
 #
 
-import threading
-import time
 import uuid
 
 from skywalking.utils.counter import AtomicCounter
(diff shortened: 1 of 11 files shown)


skywalking/trace/__init__.py
```diff
@@ -25,10 +25,13 @@
...
 class ID(object):
-    def __init__(self):
-        self.part1 = str(uuid.uuid1())
-        self.part2 = threading.get_ident()  # FIXME: thread ids are reused
-        self.part3 = int(time.time() * 1000) * 10000 + _id.next()
+    def __init__(self, string: str = None):
+        if string:
+            self.part1, self.part2, self.part3 = string.split('.')
+        else:
+            self.part1 = str(uuid.uuid1())
```
wu-sheng, line 32: If you have used UUID already, I would say, you only need one part. In java agent, the UUID is created once in the bootstrap stage.

skywalking/trace/__init__.py
```diff
@@ -25,10 +25,13 @@
 
 
 class ID(object):
-    def __init__(self):
-        self.part1 = str(uuid.uuid1())
-        self.part2 = threading.get_ident()  # FIXME: thread ids are reused
-        self.part3 = int(time.time() * 1000) * 10000 + _id.next()
+    def __init__(self, string: str = None):
+        if string:
+            self.part1, self.part2, self.part3 = string.split('.')
```
wu-sheng, line 30: Today, we don't require the 3 parts as trace id anymore. No matter string or number. Trace id/segment id are only literal string.

skywalking/trace/__init__.py
kezhenxu94: Thanks for pointing out, fixed now

PR #5, opened 2020-05-10
Feature: logging level and disabling plugins

kezhenxu94:
- Add logging level configuration item
- Add plugin disabling mechanism
- Some trivial chores


docs/FAQ.md
```diff
@@ -0,0 +1,12 @@
+## FAQ
+
+Q: How to disable some plugins?
+A: You can disable one or more plugins by following methods.
+
+```python
+from skywalking import config
+
+config.disable_plugins = ['sw_http', 'sw_request']  # can be also CSV format, i.e. 'sw_http,sw_request'
```
wu-sheng, line 9: Where should user find these names？

kezhenxu94: > Where should user find these names？

Added a section to list all the plugin names

PR #8, opened 2020-05-11
Feature: adapt HTTP protocol

kezhenxu94:
may need some performance tuning later

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -40,7 +40,7 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
       - name: Set up dependencies
-        run: make setup
+        run: make setup install
       - name: Lint codes
         run: make lint
       - name: Check license header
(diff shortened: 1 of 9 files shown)


kezhenxu94:
> No object about this feature, but the question is why? Who will need this? As clearly, gRPC provides much better performance, and the backend prefers the gRPC too.

If the user has already used gRPC in their application, and the version is incomparable with the one we are using, this would help, because there is no shaded packages in python and versions in python are somewhat sensitive

wu-sheng:
OK, get your point, but we should indicate in the document, `gRPC` is recommended. Even we don't have the comparison data, gRPC is better than HTTP, agree?

kezhenxu94:
> OK, get your point, but we should indicate in the document, `gRPC` is recommended. Even we don't have the comparison data, gRPC is better than HTTP, agree?

Sure, will add it later

wu-sheng:
Recommend add HTTP test into main repo e2e too, should be an easy work :)

PR #9, opened 2020-05-31
Add requests plugin

alonelaval:
add requests plugin

PR #11, opened 2020-06-08
Simplify the docker context path and rename test

kezhenxu94:


kezhenxu94:
This is trivial change, but good for new-comers, as what I found in #9

PR #17, opened 2020-06-17
bugfix: connected have never be called

chestarss:


chestarss:
duplicate with https://github.com/apache/skywalking-python/pull/18

PR #20, opened 2020-06-18
Add werkzeug support to sw_http_server plugin 

sungitly:
Werkzeug is widely used by python web community. Currently sw_http_server plugin doesn't support instrumenting werkzeug based python web server becasue the do_method won't be invoked at all.

See
https://github.com/pallets/werkzeug/blob/fc999a6262c847ad185ea3ffe0cc6f2e915a867a/src/werkzeug/serving.py#L328

When werkzeug WSGI handler is detected instead of BaseHTTPRequestHandler, wrap run_wsgi to add Skywalking instrument code.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -20,6 +20,9 @@ setup:
 	python3 -m pip install --upgrade pip
 	python3 -m pip install grpcio --ignore-installed
 
+setup-test: setup
+	pip3 install -e .[test]
+
 gen:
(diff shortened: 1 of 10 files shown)


kezhenxu94:
@sungitly thanks for your contribution, can you please update the corresponding test please? If you have any question on how to do that, please ping me for help

PR #22, opened 2020-06-23
Add missing step in doc

kezhenxu94:


docs/FAQ.md
```diff
@@ -20,5 +20,6 @@ that're not released yet, please clone the source codes and `make install` it:
 ```shell
 git clone https://github.com/apache/skywalking-python
 cd skywalking-python
+git submodule update --init
```
wu-sheng, line 23: I think you should correct a description above, 
> A: If you want to build the SkyWalking Python Agent from sources, for some reasons, for example, you want to try the latest features

to

> A: If you want to build the SkyWalking Python Agent from sources on the GitHub

Because we will have the Apache release(source) shortly, we don't need git-submodule there.

kezhenxu94: `from GitHub` and NB are both added

PR #25, opened 2020-06-26
Add Flask plugin

alonelaval:
Closes https://github.com/apache/skywalking/issues/4743

### DIFF
--- a/skywalking/trace/carrier/__init__.py
+++ b/skywalking/trace/carrier/__init__.py
@@ -47,7 +47,7 @@ class Carrier(CarrierItem):
         super(Carrier, self).__init__(key='sw8')
         self.trace_id = ''  # type: str
         self.segment_id = ''  # type: str
-        self.span_id = -1  # type: int
+        self.span_id = ''  # type: str
         self.service = ''  # type: str
         self.service_instance = ''  # type: str
         self.endpoint = ''  # type: str
@@ -61,7 +61,7 @@ class Carrier(CarrierItem):
             '1',
             b64encode(self.trace_id),
             b64encode(self.segment_id),
-            str(self.span_id),
+            self.span_id,
             b64encode(self.service),
             b64encode(self.service_instance),
             b64encode(self.endpoint),
@@ -74,13 +74,26 @@ class Carrier(CarrierItem):
         if not val:
             return
         parts = val.split('-')
+        if len(parts) != 8:
+            return
         self.trace_id = b64decode(parts[1])
         self.segment_id = b64decode(parts[2])
-        self.span_id = int(parts[3])
+        self.span_id = parts[3]
         self.service = b64decode(parts[4])
         self.service_instance = b64decode(parts[5])
         self.endpoint = b64decode(parts[6])
         self.client_address = b64decode(parts[7])
+
+    @property
+    def is_valid(self):
+        # type: () -> bool
+        return len(self.trace_id) > 0 and \
+               len(self.segment_id) > 0 and \
+               len(self.service) > 0 and \
+               len(self.service_instance) > 0 and \
+               len(self.endpoint) > 0 and \
+               len(self.client_address) > 0 and \
+               self.span_id.isnumeric()
 
     def __iter__(self):
         self.__iter_index = 0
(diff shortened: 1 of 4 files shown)


skywalking/plugins/sw_flask/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+            import flask
+            req = flask.request
+            context = get_context()
+            carrier = Carrier()
+            for item in carrier:
+                item.val = req.headers[item.key.capitalize()]
+            with context.new_entry_span(op=req.path, carrier=carrier) as span:
+                span.layer = Layer.Http
+                span.component = Component.Flask
+                span.peer = '%s:%s' % (req.environ["REMOTE_ADDR"], req.environ["REMOTE_PORT"])
```
kezhenxu94, line 47: Are these environment variables (`REMOTE_ADDR`, `REMOTE_PORT`) generally set when using Flask? If not, it's not appropriate to get the `peer` from these two

skywalking/plugins/sw_flask/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+                span.tag(Tag(key=tags.HttpMethod, val=req.method))
+                span.tag(Tag(key=tags.HttpUrl, val=req.url))
+                return _full_dispatch_request(this)
+
+        def _sw_handle_user_exception(this: Flask, e):
+            if e is not None:
+                entry_span = get_context().active_span()
+                entry_span.log(e)
```
kezhenxu94, line 55: We don't actually have a `log` method on `Span` type, please implement it first

kezhenxu94: And better to have a sanity check here, `entry_span` is possible to be `None`

kezhenxu94: Please also mark the span as `error` by setting `error_occurred` to `True`

kezhenxu94: And, can we get the response of the request, then we can determine the span's status by the http status code, like other plugins

skywalking/plugins/sw_flask/__init__.py
```diff
@@ -0,0 +1,63 @@
...
+            return _handle_user_exception(this, e)
+
+        Flask.full_dispatch_request = _sw_full_dispatch_request
+        Flask.handle_user_exception = _sw_handle_user_exception
+
+    except Exception:
+        logger.warning('failed to install plugin %s', __name__)
+        traceback.print_exc()
```
kezhenxu94, line 63: I suggest not to print the traceback in this plugin, other plugins are for built-in libraries, which are rarely failed to install, but this plugin is tend to be failed because not all users are using Flask

tests/plugin/sw_flask/services/consumer.py
```diff
@@ -0,0 +1,48 @@
...
+
+if __name__ == '__main__':
+    config.service_name = 'consumer'
+    config.logging_level = 'DEBUG'
+    agent.start()
+
+    import socketserver
+    from http.server import BaseHTTPRequestHandler
+
+    class SimpleHTTPRequestHandler(BaseHTTPRequestHandler):
```
kezhenxu94, line 30: It'd be better to use Flask as well in the `consumer` service, we can check the propagation of context by that

skywalking/plugins/sw_flask/__init__.py
alonelaval: yes  these environment variables  come from every request .

skywalking/plugins/sw_flask/__init__.py
alonelaval: ok

skywalking/plugins/sw_flask/__init__.py
alonelaval: ok，i use skywalking 6.6 test... i will use skywalking 8.0.1 test first.

tests/plugin/sw_flask/services/consumer.py
alonelaval: ok

wu-sheng:
Forse push and close? What happens?

PR #27, opened 2020-06-30
Add docs on how to install from Pypi

kezhenxu94:


README.md
```diff
@@ -13,6 +13,20 @@ microservices, cloud native and container-based (Docker, Kubernetes, Mesos) arch
 
 [![Build](https://github.com/apache/skywalking-python/workflows/Build/badge.svg?branch=master)](https://github.com/apache/skywalking-python/actions?query=branch%3Amaster+event%3Apush+workflow%3A%22Build%22)
 
+## Install
+
+### From Pypi (Recommended)
```
wu-sheng, line 18: I think from Apache, this should not be recommended. Because at least for now, source codes are the only official release. Someone is asking for change about this, but still, it is there.

wu-sheng: Also, I think you should add doc about install a specific version?

PR #29, opened 2020-07-01
[Core][Defect] Validate carrier before using it

kezhenxu94:


skywalking/trace/carrier/__init__.py
```diff
@@ -74,14 +74,26 @@ def val(self, val: str):
         if not val:
             return
         parts = val.split('-')
+        if len(parts) != 8:
+            return
         self.trace_id = b64decode(parts[1])
         self.segment_id = b64decode(parts[2])
-        self.span_id = int(parts[3])
+        self.span_id = int(parts[3]) if parts[3].isnumeric() else -1
```
wu-sheng, line 81: I think `isnumberic` should check in the `is_valid`

wu-sheng: Use a default -1 span could be mislinking the parent.

kezhenxu94: > Use a default -1 span could be mislinking the parent.

If it falls back to `-1`, the `is_valid` always return false, and there's no chance that the carrier is used, anyway, just update the commit to dismiss the concern

PR #30, opened 2020-07-02
[Core][BugFix] Add content equality of SegmentRef

kezhenxu94:


wu-sheng:
Is python facing duplicated trace id?

kezhenxu94:
> Is python facing duplicated trace id?

No, this fixes the duplicate refs when one framework is built on another, for example, Flask is built on top of requests

wu-sheng:
Oh, I see. So, python agent only had ref override self trace id mechanism, no ref override and ignorance mechanisms, right?

PR #32, opened 2020-07-02
[Doc] Add Flask plugin and fix requests plugin name

kezhenxu94:


README.md
```diff
@@ -63,11 +63,12 @@ Environment Variable | Description | Default
...
-| :--- | :--- |
-| [http.server](https://docs.python.org/3/library/http.server.html) | `sw_http_server` |
-| [urllib.request](https://docs.python.org/3/library/urllib.request.html) | `sw_urllib_request` |
-| [requests](https://requests.readthedocs.io/en/master/) | `requests` |
+Library | Plugin Name | Since Version
+| :--- | :--- | :--- |
+| [http.server](https://docs.python.org/3/library/http.server.html) | `sw_http_server` | v0.1.0 |
+| [urllib.request](https://docs.python.org/3/library/urllib.request.html) | `sw_urllib_request` | v0.1.0 |
+| [requests](https://requests.readthedocs.io/en/master/) | `sw_requests` | v0.1.0 |
+| [Flask](https://flask.palletsprojects.com/en/1.1.x/) | `sw_flask` | not yet released |
```
wu-sheng, line 71:
```suggestion
| [Flask](https://flask.palletsprojects.com/en/1.1.x/) | `sw_flask` | 0.2.0 |
```

How about `0.2.0`?

kezhenxu94: > How about `0.2.0`?

Because we haven't released it yet, and I don't know whether the milestone will change or not (although it's low possibility), the other problem is that users may misunderstand that there is v0.2.0 already, saying it `not yet release` may avoid unnecessary questions

wu-sheng: Actually, you don't need this version. Do you remember SkyWalking Java agent's supported list? Because we have tagged doc for different versions, we could link the documents in the README.

PR #33, opened 2020-07-05
Add component ID for requests plugin

alonelaval:
add Component Id

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,6 +6,5 @@
 __pycache__
 build/
 dist/
-skywalking_python.egg-info/
 *.pyc
 /apache_skywalking.egg-info/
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -7,4 +7,5 @@ __pycache__
 build/
 dist/
 skywalking_python.egg-info/
-*.pyc
\ No newline at end of file
+*.pyc
+/apache_skywalking.egg-info/
```
kezhenxu94: Please remove the `skywalking_python.egg-info` (at line 9), I've forgot to rename it after I rename the module name, thanks

alonelaval: ok

PR #36, opened 2020-07-16
Refactor: move __init__.py to individual file

kezhenxu94:


wu-sheng:
Seems to have the conflict right now. Please fix.
BTW, what is the expected 0.2.0 release date?

kezhenxu94:
The 0.2.0 is due by then end of this month, July 30, 2020

PR #37, opened 2020-07-16
Add plugin for Django Agent

Humbertzhang:
Closes https://github.com/apache/skywalking/issues/4744


skywalking/plugins/sw_django/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+                if entry_span is not None and type(entry_span) is not NoopSpan:
+                    entry_span.raised()
+
+            return _handle_uncaught_exception(request, resolver, exc_info)
+
+        BaseHandler.get_response = _sw_get_response
+        exception.handle_uncaught_exception = _sw_handle_uncaught_exception
+    except Exception:
+        logger.warning('failed to install plugin %s', __name__)
+        traceback.print_exc()
```
kezhenxu94, line 75: Let's not print the traceback, because it is high possible that the users don't use Django, printing such traceback is a panic for them

skywalking/plugins/sw_django/__init__.py
```diff
@@ -0,0 +1,75 @@
...
+                span.tag(Tag(key=tags.HttpStatus, val=resp.status_code))
+                if resp.status_code >= 400:
+                    span.error_occurred = True
+
+                return resp
+
+        def _sw_handle_uncaught_exception(request, resolver, exc_info):
+            if exc_info is not None:
+                entry_span = get_context().active_span()
+                if entry_span is not None and type(entry_span) is not NoopSpan:
```
kezhenxu94, line 66:
```suggestion
                if entry_span is not None and not isinstance(entry_span, NoopSpan):
```

By the way, calling `raised` on `NoopSpan` has no side effects, so is it necessary to have the condition `not isinstance(entry_span, NoopSpan)`?

Humbertzhang: I see, i will just check if a span is None and no more check if it is a NoopSpan.

skywalking/plugins/sw_django/__init__.py
Humbertzhang: Got it, you are right.

kezhenxu94:
@Humbertzhang can you please also add the component id to the main repo, and add a logo to the UI repo?

UI REPO: http://github.com/apache/skywalking-rocketbot-ui
MAIN REPO: http://github.com/apache/skywalking

Humbertzhang:
okay, i will open pr for that.

kezhenxu94:
Another enhancement is to add an option to support collecting the http request parameters, if you're interested in :-)

kezhenxu94:
@Humbertzhang the plugin is missed in the doc, can you add it to the doc please? 

https://github.com/apache/skywalking-python/blob/master/README.md#supported-libraries

Humbertzhang:
ok, i will add this plugin in README.md. 
And I will figure out what i should do for collecting the http request parameters, and i will try my best to do it.

PR #38, opened 2020-07-16
Created venv builders for linux/windows and req flashers + use documentation

zkscpqm:
Scripts folder to hold auxilliary, non-application helpers.

Initial contents:

build_requirements (win/linux): This will have a look at your virtual environment packages and flash them to a "requirements.txt" file. This file will then allow developers to set up their virtual environment easily

setup (win/linux): This will create a separate clean virtual environment for the project and only the necessary packages (outlined in requirements.txt) will be installed. 

Updated the gitignore file to ignore the virtual environment

Updated README file with instructions on building and updating the requirements


zkscpqm:
> The newly-added files require a license header, you can copy one from the existing files

Done, but as one of my files is a batch file, the comments use `::` instead of `#` so your license checker script will probably pick that up

kezhenxu94:
> > The newly-added files require a license header, you can copy one from the existing files
> 
> Done, but as one of my files is a batch file, the comments use `::` instead of `#` so your license checker script will probably pick that up

NO worries, ATM I just check 3 directories `skywalking, tests, tools`, so the aforementioned review comments are made manually, and it won't fail, we can add more directories to the check list if needed

kezhenxu94:
Thanks, I'll take a detailed look later

kezhenxu94:
@zkscpqm do we have a corresponding script for Mac OS?

zkscpqm:
> @zkscpqm do we have a corresponding script for Mac OS?

I'd like to say that the Linux scripts will work for MacOS too, but I don't have access to a Mac Environment to test. The shebang is set to `#!/bin/sh` and I'm not sure if that's where the binary for a Mac shell is. If you have access to a device running Mac OS, give it a go.

kezhenxu94:
> > @zkscpqm do we have a corresponding script for Mac OS?
> 
> I'd like to say that the Linux scripts will work for MacOS too, but I don't have access to a Mac Environment to test. The shebang is set to `#!/bin/sh` and I'm not sure if that's where the binary for a Mac shell is. If you have access to a device running Mac OS, give it a go.

Sure, that's what I mean because I'm trying to test it locally (with a Mac), I'll test with the Linux script then

README.md
```diff
@@ -127,5 +127,22 @@ with context.new_entry_span(op=str('https://github.com/apache/skywalking')) as s
 
 Check [the FAQ page](docs/FAQ.md) or add the FAQs there.
 
+## For Developers
+
+### Steps to get an operational virtual environment:
+
+1. `git clone https://github.com/apache/skywalking-python.git`
+2. `cd skywalking-python/scripts` (*make sure you actually go into the directory since the scripts use relative paths*)
```
kezhenxu94, line 135: nit: I hope the script itself can tackle the paths by some extra work, that would be more friendly to the developers

README.md
```diff
@@ -127,5 +127,22 @@ with context.new_entry_span(op=str('https://github.com/apache/skywalking')) as s
...
+3. Run the script for your relevant OS to create a virtual environment folder in the project root (*skywalking-python/venv*) and install all the necessary requirements
+
+**Make sure that when the `python` command is executed on your workstation, the binary it references is python 3.5 or newer!**
```
kezhenxu94, line 138: IIRC this can be automatically determined right?

scripts/build_requirements_linux.sh
```diff
@@ -0,0 +1,24 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+#!/bin/sh
+
+echo "Building requirements.txt"
+
+../venv/Scripts/python -m pip freeze > ../requirements.txt
```
kezhenxu94, line 22: The path under Mac OS is `../venv/bin/python` instead of `../venv/Scripts/python`

scripts/setup_linux.sh
```diff
@@ -0,0 +1,29 @@
...
+# limitations under the License.
+#
+
+#!/bin/sh
+
+echo "Creating virtual environment"
+python -m venv '../venv'
+echo "Virtual env created"
+
+../venv/Scripts/python.exe -m pip install --upgrade pip
```
kezhenxu94, line 24: ditto, and `.exe` suffix should be removed

scripts/setup_linux.sh
```diff
@@ -0,0 +1,29 @@
...
+echo "Pip upgrade complete. Installing packages from requirements.txt"
+
+while read requirement; do
+  ../venv/Scripts/python.exe -m pip install $requirement
+done < ../requirements.txt
```
kezhenxu94, line 29: Can be replaced by `xxx install -r ../requirements.txt`, and no `.exe` suffix

scripts/setup_windows.ps1
```diff
@@ -0,0 +1,31 @@
...
+Write-Output "Virtual env created"
+
+$PYEXE= Join-Path -Path $PSScriptRoot -ChildPath '..\venv\Scripts\python.exe'
+& $PYEXE -m pip install --upgrade pip
+
+Write-Output "Pip upgrade complete. Installing packages from requirements.txt"
+
+foreach($package in Get-Content ..\requirements.txt) {
+    & $PYEXE -m pip install $package
+}
```
kezhenxu94, line 31: ditto

README.md
zkscpqm: I'll have to double-check but I think you can determine either python3 or perhaps python37 but IDK about specifically python35-python38. And sometimes python3 is not implemented as an option (on my PC, writing python3 in PowerShell redirects me to the Microsoft Store :D

requirements.txt
```diff
@@ -0,0 +1,23 @@
...
+pyparsing==2.4.7
+pypiwin32==223
+pywin32==228
+requests==2.24.0
+six==1.15.0
+testcontainers==3.0.3
+urllib3==1.25.9
+websocket-client==0.57.0
+Werkzeug==1.0.1
+wrapt==1.12.1
```
kezhenxu94, line 23: Do you think that we should check in this file? I saw this is Windows-specific, and in my side, it should be Mac-specific, this file keeps changing from developer to developer

kezhenxu94: And it seems to be created automatically in the script, so don't need to check in

zkscpqm: The point of this file is for the build script to work. If you don't have this file, the other 4 scripts are pointless. It tells them what libraries need to be installed in the virtual env so that the dev doesn't have to go file-to-file to see what they need to install

zkscpqm: If we don't check it in, a new developer/user who is just downloading the source will not have a way to set up their venv

kezhenxu94: According to your doc, if the developers can `2. Execute the build_requirements script relevant to your OS.` to generate the `requirements.txt`, why should we check the `requirements.txt` in? Other developers can also execute the same script to generate it, right?

kezhenxu94:
@zkscpqm and I found a dead lock in your scripts, `build_requirements_linux.sh` requires `venv`, but `venv` is set up in `setup_linux.sh`, while `setup_linux.sh` install dependencies from `requirements.txt` which is generated in `build_requirements_linux.sh`, please re-consider this

PR #39, opened 2020-07-17
Add PyMsql Sql Parameters

alonelaval:


kezhenxu94:
@alonelaval can you add the config into the existing test case and verify?

alonelaval:
> @alonelaval can you add the config into the existing test case and verify?

ok

PR #43, opened 2020-07-19
Add Flask Http Params

alonelaval:


alonelaval:
> Please also add the config to the doc

ok

PR #44, opened 2020-07-20
Add Redis Plugin

alonelaval:


alonelaval:
> So cool, LGTM, just a reminder, add a component id and mapping to the main repo, no logo is needed, you may be already familiar with these :)

OK

PR #53, opened 2020-07-26
Add Rabbitmq Plugin

alonelaval:
resolve apache/skywalking#5172


wu-sheng:
Please use resolves https://github.com/apache/skywalking/issues/5172 to link the PR. And only close the issue after PR merged.

alonelaval:
> Please use resolves [apache/skywalking#5172](https://github.com/apache/skywalking/issues/5172) to link the PR. And only close the issue after PR merged.

ok

PR #55, opened 2020-07-31
Add CorrelationContext

alonelaval:
resolve apache/skywalking#5189

### DIFF
--- a/skywalking/trace/context/__init__.py
+++ b/skywalking/trace/context/__init__.py
@@ -125,7 +125,10 @@ class SpanContext(object):
         return None
 
     def put_correlation(self, key, value):
-        if key is None or value is None:
+        if key is None:
+            return
+        if value is None:
+            self._correlation.pop(key, value)
             return
         if len(value) > config.correlation_value_max_length:
             return
--- a/skywalking/trace/context/__init__.py
+++ b/skywalking/trace/context/__init__.py
@@ -33,6 +33,7 @@ class SpanContext(object):
         self.spans = []  # type: List[Span]
         self.segment = Segment()  # type: Segment
         self._sid = Counter()
+        self._correlation = {}  # type: dict
 
     def new_local_span(self, op: str) -> Span:
         span = self.ignore_check(op, Kind.Local)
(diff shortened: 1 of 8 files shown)


tests/plugin/sw_flask/services/provider.py
```diff
@@ -24,12 +24,16 @@
...
+    from flask import Flask, jsonify, request
 
     app = Flask(__name__)
 
     @app.route("/users", methods=["POST", "GET"])
     def application():
+        print(request.headers)
+        from skywalking.trace.context import get_context
+        print(get_context().get_correlation("test"))
+        print(get_context().get_correlation("test2"))
```
kezhenxu94, line 36: Do not simply print the correlation headers, this can not verify anything, even the headers are empty, the tests still passed, one way is to  propagate these headers to the test codes, return the headers as the http response, for example, then get the response and verify the expected data here, `self.validate()` returns the response, and you can get the response and verify further

https://github.com/apache/skywalking-python/blob/0246634b8e908b845e1540c6d2b70d9058794601/tests/plugin/sw_flask/test_flask.py#L38

skywalking/trace/context/__init__.py
```diff
@@ -118,23 +119,44 @@ def active_span(self):
 
         return None
 
+    def get_correlation(self, key):
+        if key in self._correlation:
+            return self._correlation[key]
+        return None
+
+    def put_correlation(self, key, value):
+        if key is None or value is None:
```
kezhenxu94, line 128: If the `value is None`, I'm expecting the users want to remove the item, but you just ignore it

skywalking/trace/span/__init__.py
```diff
@@ -225,5 +226,10 @@ class NoopSpan(Span):
     def __init__(self, context: 'SpanContext' = None, kind: 'Kind' = None):
         Span.__init__(self, context=context, kind=kind)
 
+    def extract(self, carrier: 'Carrier') -> 'Span':
+        if carrier is not None:
+            self.context._correlation = carrier.correlation_carrier.correlation
+
     def inject(self, carrier: 'Carrier') -> 'Span':
+        carrier.correlation_carrier.correlation = self.context._correlation
```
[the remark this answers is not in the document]
alonelaval: I want to protect context._correlation and let the user get or set values that need to be passed through through put and get

skywalking/trace/context/__init__.py
alonelaval: ok

tests/plugin/sw_flask/services/provider.py
alonelaval: ok

PR #58, opened 2020-08-06
[Doc] Add dev doc and reorganize the structure

kezhenxu94:
closes https://github.com/apache/skywalking/issues/4993


kezhenxu94:
> `requirements.txt` has a lot of changes, but no LICENSE update, is that expected?

Expected, they're missing by the previous contributor, and all the dependencies in requirements.txt are **development only**, not in distributed package or runtime

PR #63, opened 2020-08-11
[Plugin] check version of packages when install plugins

Humbertzhang:

### DIFF
--- a/skywalking/plugins/__init__.py
+++ b/skywalking/plugins/__init__.py
@@ -42,9 +42,10 @@ def install():
         logger.debug('installing plugin %s', modname)
         plugin = importer.find_module(modname).load_module(modname)
 
-        logger.debug('checking version for plugin %s', modname)
-        supported = pkg_version_check(plugin, modname)
+        supported = pkg_version_check(plugin)
         if not supported:
+            logger.debug('check version for plugin %s\'s corresponding package failed, thus '
+                         'won\'t be installed', modname)
             continue
 
         if not hasattr(plugin, 'install') or inspect.ismethod(getattr(plugin, 'install')):
@@ -56,14 +57,19 @@ def install():
 _operators = {
     '<': lambda cv, ev: cv < ev,
     '<=': lambda cv, ev: cv < ev or cv == ev,
-    '=': lambda cv, ev: cv == ev,
+    '==': lambda cv, ev: cv == ev,
     '>=': lambda cv, ev: cv > ev or cv == ev,
     '>': lambda cv, ev: cv > ev,
     '!=': lambda cv, ev: cv != ev
 }
 
 
-def pkg_version_check(plugin, modname):
+class VersionRuleException(Exception):
+    def __init__(self, message):
+        self.message = message
+
+
+def pkg_version_check(plugin):
     supported = True
 
     # no version rules was set, no checks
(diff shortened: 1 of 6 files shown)


skywalking/plugins/__init__.py
```diff
@@ -38,7 +41,64 @@ def install():
...
+    '<': lambda cv, ev: cv < ev,
+    '<=': lambda cv, ev: cv < ev or cv == ev,
+    '=': lambda cv, ev: cv == ev,
+    '>=': lambda cv, ev: cv > ev or cv == ev,
+    '>': lambda cv, ev: cv > ev,
+    '!=': lambda cv, ev: cv != ev
+}
+
+
+def pkg_version_check(plugin, modname):
```
kezhenxu94, line 65: Can you add some unit tests for this function? Just to check the `operators` work correctly

skywalking/plugins/__init__.py
```diff
@@ -38,7 +41,64 @@ def install():
             continue
         logger.debug('installing plugin %s', modname)
         plugin = importer.find_module(modname).load_module(modname)
+
+        ok = pkg_version_check(plugin, modname)
+        if not ok:
```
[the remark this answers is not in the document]
Humbertzhang: Is it ok to log `logger.debug('checking version for plugin %s', modname)` ?

skywalking/plugins/__init__.py
```diff
@@ -38,7 +41,65 @@ def install():
             continue
         logger.debug('installing plugin %s', modname)
         plugin = importer.find_module(modname).load_module(modname)
+
+        logger.debug('checking version for plugin %s', modname)
+        supported = pkg_version_check(plugin, modname)
+        if not supported:
+            continue
```
kezhenxu94, line 48: [1] move `logger.debug('checking version for plugin %s', modname)` into the `if not supported` block, and reword the message to tell the users that the current version is not supported by the plugin

skywalking/plugins/__init__.py
```diff
@@ -38,7 +41,65 @@ def install():
...
+
+    pkg_name = plugin.version_rule.get("name")
+    rules = plugin.version_rule.get("rules")
+
+    try:
+        current_pkg_version = pkg_resources.get_distribution(pkg_name).version
+    except pkg_resources.DistributionNotFound:
+        supported = False
+        logger.warning("plugin %s didn\'t find the corresponding package %s, thus won't be installed",
+                       modname, pkg_name)
```
kezhenxu94, line 81: When failed to get the version, we can just consider it as supported as the built-in libraries should always throw exception here

skywalking/plugins/__init__.py
```diff
@@ -38,7 +41,65 @@ def install():
...
+        expect_pkg_version = rule[idx:]
+
+        current_version = version.parse(current_pkg_version)
+        expect_version = version.parse(expect_pkg_version)
+        f = _operators.get(symbol) or None
+
+        # version rule parse error, take it as no more version rules and return True.
+        if not f:
+            logger.warning("plugin %s version rule %s error. only allow >,>=,=,<=,<,!= symbols", modname, rule)
+            return supported
```
kezhenxu94, line 97: We can simply raise an exception and exit immediately, this kind of exception should be fixed at development phase and never happened at runtime

skywalking/plugins/__init__.py
```diff
@@ -38,7 +41,65 @@ def install():
...
+
+        if not f(current_version, expect_version):
+            supported = False
+            logger.warning("plugin %s need package %s version follow rules %s ,current version " +
+                           "is %s, thus won\'t be installed", modname, pkg_name, str(rules), current_pkg_version)
```
kezhenxu94, line 102: Move this kind of message to where the invoker calls this method, I mark [1] in the previous comment

image

PR #75, opened 2020-09-30
Fixed issues with shebang

championquizzer:
Fixed issues with the placement of shebang in two files.


kezhenxu94:
I'm sorry I just merged https://github.com/apache/skywalking-python/pull/77 without noticing this one is earlier than that one, anyway thanks for opening the PR :)

PR #76, opened 2020-09-30
Mitigate unsafe yaml loading

langyizhao:
This is to mitigate security issue `Use of unsafe yaml load` found in https://console.muse.dev/result/ayorra/skywalking-python/01EK1BP85DHDEV2RJTTEYFNV9G?tab=results

According to https://pyyaml.org/wiki/PyYAMLDocumentation
Ideally we should always use safe_load in place of load when the source of yaml can be untrusted (which IS the case because it's from the web)

But I understand that we also want to use CLoader for the sake of performance.
This is a mitigation following https://github.com/yaml/pyyaml/blob/2f463cf5b0e98a52bc20e348d1e69761bf263b86/tests/lib/test_yaml_ext.py#L37

### DIFF
--- a/tests/plugin/base.py
+++ b/tests/plugin/base.py
@@ -30,7 +30,7 @@ from requests import Response
 try:
     from yaml import CSafeLoader as Loader
 except ImportError:
-    from yaml import SafeLoader
+    from yaml import SafeLoader as Loader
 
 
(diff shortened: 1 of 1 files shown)


kezhenxu94 (approved):
Test codes only, LGTM

PR #79, opened 2020-11-07
bugfix: agent cannot reconnect after server is down

kezhenxu94:
Fixes https://github.com/apache/skywalking/issues/5732

### DIFF
--- a/skywalking/agent/protocol/grpc.py
+++ b/skywalking/agent/protocol/grpc.py
@@ -16,6 +16,7 @@
 #
 
 import logging
+import traceback
 from queue import Queue
 
 import grpc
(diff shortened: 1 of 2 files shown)


skywalking/agent/protocol/grpc.py
```diff
@@ -40,22 +41,33 @@ def __init__(self):
...
+        except grpc.RpcError:
+            self.on_error()
 
     def connected(self):
         return self.state == grpc.ChannelConnectivity.READY
 
+    def on_error(self):
+        traceback.print_exc()
+        self.channel.unsubscribe(self._cb)
+        self.channel.subscribe(self._cb, try_to_connect=True)
```
kezhenxu94, line 69: The reason why it didn't reconnect before, is that the `try_to_connect` only takes effect in this single subscription, once the channel is connected, it is set to `False`, therefore we re-subscribe once the server closes this connection.

From https://github.com/grpc/grpc/issues/6478#issuecomment-226951042

wu-sheng (approved):

PR #81, opened 2020-11-19
Bump up the version to 0.4.0

kezhenxu94:
Forgot to update the version number before voting 😢 @wu-sheng do we need another round voting? I hope no

Can I update this version number (and re-tag) and upload to Pypi? (Can't upload without updating the version)

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -25,7 +25,7 @@ README = (HERE / "README.md").read_text()
 
 setup(
     name="apache-skywalking",
-    version="0.3.0",
+    version="0.4.0",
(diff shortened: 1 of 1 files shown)


wu-sheng (approved):
I am afraid you have to re-vote the release, as you changed the commit id sadly.

wu-sheng:
I am assuming this file is in the release tar ball, right?

kezhenxu94:
> I am assuming this file is in the release tar ball, right?

Yes it is

wu-sheng:
Then we need another release vote.

PR #83, opened 2020-11-19
Refactor: moved exception check to Span.__exit__

tom-pytel:
Moved exception check for spans into Span.__exit__ from multiple individual locations in individual plugins. This should also catch other errors that may not be recognized otherwise. Is there any reason not to do it like this?

### DIFF
--- a/skywalking/trace/span.py
+++ b/skywalking/trace/span.py
@@ -117,6 +117,8 @@ class Span(ABC):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
+        if isinstance(exc_type, BaseException):
+            self.raised()
         self.stop()
         if exc_tb is not None:
             return False
--- a/skywalking/plugins/sw_elasticsearch.py
+++ b/skywalking/plugins/sw_elasticsearch.py
@@ -34,16 +34,12 @@ def install():
         with context.new_exit_span(op="Elasticsearch/" + method + url, peer=peer) as span:
             span.layer = Layer.Database
             span.component = Component.Elasticsearch
-            try:
-                res = _perform_request(this, method, url, headers=headers, params=params, body=body)
+            res = _perform_request(this, method, url, headers=headers, params=params, body=body)
 
-                span.tag(Tag(key=tags.DbType, val="Elasticsearch"))
-                if config.elasticsearch_trace_dsl:
-                    span.tag(Tag(key=tags.DbStatement, val="" if body is None else body))
+            span.tag(Tag(key=tags.DbType, val="Elasticsearch"))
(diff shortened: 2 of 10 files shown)


skywalking/trace/span.py
```diff
@@ -117,6 +117,8 @@ def __enter__(self):
         return self
 
     def __exit__(self, exc_type, exc_val, exc_tb):
+        if isinstance(exc_type, BaseException):
+            self.raised()
```
kezhenxu94, line 121: Will this stop the propagation of the exception? We'd rather not to swallow the exceptions from the original codes, that's why we need to raise them again after we captured them, like this

```python
            try:
                 res = _open(this, fullurl, data, timeout)
                 span.tag(Tag(key=tags.HttpMethod, val=fullurl.get_method()))
                 span.tag(Tag(key=tags.HttpUrl, val=fullurl.full_url))
                 span.tag(Tag(key=tags.HttpStatus, val=res.code))
                 if res.code >= 400:
                     span.error_occurred = True
             except HTTPError as e:
                 span.raised()
                 raise e  # <- this one
```

kezhenxu94 (commented):
Hey @tom-pytel, thanks very much for noticing this, I have only one question inline

skywalking/trace/span.py
tom-pytel: No this will not stop propagation.

kezhenxu94: > No this will not stop propagation.

Thanks for clarifying, LGTM

kezhenxu94 (commented):
@tom-pytel could you please kindly refactor the counterpart in `context.new_entry_span` as well? Now you just refactored the `context.new_exit_span`, really appreciate 🙇

tom-pytel:
> @tom-pytel could you please kindly refactor the counterpart in `context.new_entry_span` as well? Now you just refactored the `context.new_exit_span`, really appreciate 🙇
> 
> e.g.
> 
> https://github.com/apache/skywalking-python/blob/3b74010b150f017e3f4fa20e1ba910ed913efb6a/skywalking/plugins/sw_rabbitmq.py#L94-L101

The exception checking is in class `Span` which is base for `EntrySpan`, `ExitSpan` and `NoopSpan` so it should work for all.

kezhenxu94:
> The exception checking is in class `Span` which is base for `EntrySpan`, `ExitSpan` and `NoopSpan` so it should work for all.

I meant removing the `try-catch` in the `new_entry_span` block, I missed that you already removed them 😄 thanks

kezhenxu94 (approved):
Thanks

tom-pytel:
> Thanks

No problem 😃

PR #84, opened 2020-11-19
[Plugin] moved Method & Url tags before requests

tom-pytel:
* Moved set of HttpMethod and HttpUrl tags to before requests to make sure they are recorded even if request fails.
* Specifically catching HTTPError in `sw_urllib_request.py` to properly set HttpStatus in this case.

### DIFF
--- a/skywalking/plugins/sw_urllib_request.py
+++ b/skywalking/plugins/sw_urllib_request.py
@@ -46,14 +46,14 @@ def install():
 
             [fullurl.add_header(item.key, item.val) for item in carrier]
 
-            span.tag(Tag(key=tags.HttpMethod, val=fullurl.get_method()))
-            span.tag(Tag(key=tags.HttpUrl, val=fullurl.full_url))
-
             try:
                 res = _open(this, fullurl, data, timeout)
             except HTTPError as e:
                 span.tag(Tag(key=tags.HttpStatus, val=e.code))
                 raise
+            finally:  # we do this here because it may change in _open()
+                span.tag(Tag(key=tags.HttpMethod, val=fullurl.get_method()))
+                span.tag(Tag(key=tags.HttpUrl, val=fullurl.full_url))
 
             span.tag(Tag(key=tags.HttpStatus, val=res.code))
             if res.code >= 400:
--- a/skywalking/plugins/sw_urllib_request.py
+++ b/skywalking/plugins/sw_urllib_request.py
@@ -29,6 +29,7 @@ logger = logging.getLogger(__name__)
 
 def install():
     from urllib.request import OpenerDirector
+    from urllib.error import HTTPError
 
     _open = OpenerDirector.open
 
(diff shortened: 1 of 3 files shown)


tom-pytel:
I expect these changes to change some test results, adding HttpMethod and HttpUrl tags in places where they may not have been before and changing some endpoint names (removing args following ?), any way to see if these test failures are that or something else?

kezhenxu94:
> any way to see if these test failures are that or something else?

Here it is https://github.com/apache/skywalking-python/pull/84/checks?check_run_id=1426036106

tom-pytel:
> > any way to see if these test failures are that or something else?
> 
> Here it is https://github.com/apache/skywalking-python/pull/84/checks?check_run_id=1426036106
> 
> Pay attention to the diff in the log

Yes I did see that but I am not familiar with the test so I don't know if the diff means that HttpMethod and HttpUrl tags appeared where they were not before or if it means some other failure mode.

skywalking/plugins/sw_urllib_request.py
```diff
@@ -38,15 +39,22 @@ def _sw_open(this: OpenerDirector, fullurl, data, timeout):
...
             span.component = Component.General
 
             [fullurl.add_header(item.key, item.val) for item in carrier]
 
-            res = _open(this, fullurl, data, timeout)
             span.tag(Tag(key=tags.HttpMethod, val=fullurl.get_method()))
             span.tag(Tag(key=tags.HttpUrl, val=fullurl.full_url))
+
+            try:
+                res = _open(this, fullurl, data, timeout)
```
kezhenxu94, line 53: I think this breaks the tests, `fullurl.get_method()` is not static, it differs before/after `_open(this, fullurl, data, timeout)`

skywalking/plugins/sw_urllib_request.py
```diff
@@ -38,15 +39,23 @@ def _sw_open(this: OpenerDirector, fullurl, data, timeout):
...
-            span.tag(Tag(key=tags.HttpMethod, val=fullurl.get_method()))
-            span.tag(Tag(key=tags.HttpUrl, val=fullurl.full_url))
+            try:
+                res = _open(this, fullurl, data, timeout)
+                span.tag(Tag(key=tags.HttpMethod, val=fullurl.get_method()))
+            except HTTPError as e:
+                span.tag(Tag(key=tags.HttpStatus, val=e.code))
+                raise
+            finally:  # we do this here because it may change in _open()
+                span.tag(Tag(key=tags.HttpMethod, val=fullurl.get_method()))
```
kezhenxu94, line 56: Please note that tag with the same key can co-exist, here you tag the span twice with the same key (`tags.HttpMethod`), which breaks the tests

tom-pytel: You are right I forgot to remove that, but I don't think that is the source of the problem, lets see...

kezhenxu94: > You are right I forgot to remove that, but I don't think that is the source of the problem (or at least the only problem), lets see...

What potential problem do you foresee? From the code level, I think it looks good. As the test case is simple (endpoint names without `?`), so the changes should be OK IMO

tom-pytel: Well, because this does change behavior. If there are tests that previously did not expect these tags due to testing a fail case and they are now returned then that will be a difference.

kezhenxu94: > Well, because this does change behavior. If there are tests that previously did not expect these tags due to testing a fail case and they are now returned then that will be a difference.

OK, that makes sense, IIRC, we didn't test fail case, so the tests should pass

kezhenxu94 (approved):
Thanks

PR #85, opened 2020-11-20
[Plugin] http servers set URL and status tags

tom-pytel:
* Base HTTP and werkzeug servers instrumented to store status code when that is sent and set it as tag for active span, as well as URL from headers.
* Fixed Span.tag() - Previously would try to change an immutable property when overriding, now replaces entire Tag.
* Fixed sw_urllib_request.py:sw_open() - Added default values to parameters because without them it would fail when called internally in urllib.request without a `data` argument, like in `HTTPRedirectHandler.http_error_302()`.

### DIFF
--- a/skywalking/plugins/sw_http_server.py
+++ b/skywalking/plugins/sw_http_server.py
@@ -41,6 +41,14 @@ def install():
 
     BaseHTTPRequestHandler.handle = _sw_handle
 
+    def _sw_send_response_only(self, code, *args, **kwargs):
+        self._status_code = code
+
+        return _send_response_only(self, code, *args, **kwargs)
+
+    _send_response_only = BaseHTTPRequestHandler.send_response_only
+    BaseHTTPRequestHandler.send_response_only = _sw_send_response_only
+
 
 def wrap_werkzeug_request_handler(handler):
     """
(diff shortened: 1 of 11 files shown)


skywalking/plugins/sw_http_server.py
```diff
@@ -53,15 +61,38 @@ def _wrap_run_wsgi():
...
-        with context.new_entry_span(op=handler.path, carrier=carrier) as span:
+        path = handler.path or '/'
+        with context.new_entry_span(op=path.split("?")[0], carrier=carrier) as span:
+            url = 'http://' + handler.headers["Host"] + path if 'Host' in handler.headers else path
             span.layer = Layer.Http
             span.component = Component.General
             span.peer = '%s:%s' % handler.client_address
             span.tag(Tag(key=tags.HttpMethod, val=handler.command))
-            return _run_wsgi()
+            span.tag(Tag(key=tags.HttpUrl, val=url))
```
kezhenxu94, line 71: I'm afraid that this will break the tests (not sure though, didn't check the tests), you added a new tag, so you may need to adjust the expected data file to match it.

> As you're actively contributing ❤️ , I hope you can set up the test environment locally so that you can quickly verify the part you modified (if you need any help, feel free to ping me anytime on Slack or GitHub), no need to wait for the GitHub Actions (GHA) for a long time, 'cause it runs all the tests

kezhenxu94: Basically, we only need docker (and docker-compose), and you can run one or multiple tests under https://github.com/apache/skywalking-python/tree/master/tests just like you run a unit test in Python, e.g. `pytest -v tests/plugin/sw_http`

tom-pytel: That is what I meant with the previous PR that the tags changed, but in that case only failing cases would change, with this change normal tests will change so yes they may need to be updated to expect the new tag.

Maybe will look into setting up test env but don't have too much more to add, maybe an sw_wsgiref plugin but will see, so for now lets see how this test comes back and hope its an easy fix.

kezhenxu94: > Maybe will look into setting up test env but don't have too much more to add, maybe an sw_wsgiref plugin but will see, so for now lets see how this test comes back and hope its an easy fix.

OK, so the there are diffs in test results, if you don't want to set up the env locally, they should be helpful to adjust, or, if you need any help, I can debug locally for you

kezhenxu94: image

Just FYI, the circled is what need to be adjusted, and ignore the `isError` part, and something like `gt 0` is for verifying uncertain integer (timestamp in this case)

tom-pytel: Yeah I looked into it and made the changes, this latest push should succeed. Also noticed why the previous PR didn't have problems, there are no tests for fail cases? 404? 500? There really should be...

tom-pytel:
This last commit is the following - Fixed sw_urllib_request.py:sw_open() - Added default values to parameters because without them it would fail when called internally in urllib.request without a `data` argument, like in `HTTPRedirectHandler.http_error_302()`. Also a minor fix for a silly oversight in my previous `Span.__exit__()` change.

kezhenxu94 (approved):
Thanks very much

PR #86, opened 2020-11-22
[Enhancement] Optimized fast_path_match()

tom-pytel:
* Changed fast_path_match() to use Python regex instead of working one character at a time.
* Also pulled out invariant disable_patterns out of loop in plugins.install().

### DIFF
--- a/skywalking/utils/ant_matcher.py
+++ b/skywalking/utils/ant_matcher.py
@@ -16,75 +16,24 @@
 #
 
 
+import re
+
+reesc = re.compile(r'([.*+?^=!:${}()|\[\]\\])')
+recache = {}
+
+
 def fast_path_match(pattern: str, path: str):
-    return normal_match(pattern, 0, path, 0)
+    repat = recache.get(pattern)
 
+    if repat is None:
+        repat = recache[pattern] = \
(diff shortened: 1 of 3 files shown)


skywalking/utils/ant_matcher.py
```diff
@@ -16,75 +16,24 @@
 #
 
 
-def fast_path_match(pattern: str, path: str):
```
kezhenxu94, line 19: This is ported from a Java version, I've just run a simple benchmark test, regex is much more better then the original version, thanks

kezhenxu94 (approved):
LGTM

tom-pytel:
Hey, can't seem to find a way to open an issue on the NodeJS agent so will ask here, having a problem running the test due to the following:
```
Step 1/5 : FROM nodejs-dependencies
    ✓ test /home/tom/src/revdebugjs/skywalking-nodejs/tests/plugins/http
Service 'server' failed to build: pull access denied for nodejs-dependencies, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```
Any idea?

wu-sheng:
Hi @tom-pytel we share the issue list in the main repo to make people easier, feel free to submit a discussion here, https://github.com/apache/skywalking/issues with project name as the prefix of the issue title. :)

PR #87, opened 2020-11-22
Fix: Non-global skywalking logging

tom-pytel:
* Changed logging to use a single skywalking Logger instance instead of global Logger and settings to avoid changing or breaking  host application logging with skywalking logger settings.
* Also added custom logging level "OFF".

### DIFF
--- a/skywalking/loggings.py
+++ b/skywalking/loggings.py
@@ -34,5 +34,5 @@ logger = getLogger('skywalking')
 
 
 def init():
-    logging.addLevelName(logging.getLevelName('CRITICAL') + 10, 'OFF')
+    logging.addLevelName(logging.CRITICAL + 10, 'OFF')
     logger.setLevel(logging.getLevelName(config.logging_level))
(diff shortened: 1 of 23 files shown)


skywalking/loggings.py
```diff
@@ -20,8 +20,19 @@
...
+
+logger = getLogger('skywalking')
+
+
 def init():
-    logging.basicConfig(
-        level=logging.getLevelName(config.logging_level),
-        format='%(name)-32s [%(threadName)-15s] [%(levelname)-8s] %(message)s',
-    )
+    logging.addLevelName(logging.getLevelName('CRITICAL') + 10, 'OFF')
```
kezhenxu94, line 37:
```suggestion
    logging.addLevelName(logging.CRITICAL + 10, 'OFF')
```

kezhenxu94 (commented):
Look good basically, except for one nit

skywalking/loggings.py
tom-pytel: Sure, if you want to do it yourself go for it, if not I will do it tomorrow.

kezhenxu94 (approved):

PR #88, opened 2020-11-23
[Enhancement] Async task context support

tom-pytel:
* Async task context usage for Python 3.7+ for things like Tornado.
* Also grpc reconnect traceback print only in DEBUG mode.

This will allow async frameworks and apps to mostly work properly - where with only thread-local storage they could not as the individual async tasks would interfere with each other's context. A new task-local context is created on get_context() for each task and all subtasks inherit this to be able to correlate entry and exit spans.

This is not a complete and final solution as there will still be problems with a single entry point executing multiple async exit requests simultaneously. Also if a context is created BEFORE the async handler loop starts then all the handlers would share the same parent context, but this does not happen in Tornado.

### DIFF
--- a/skywalking/agent/protocol/grpc.py
+++ b/skywalking/agent/protocol/grpc.py
@@ -15,6 +15,7 @@
 # limitations under the License.
 #
 
+import logging
 from skywalking.loggings import logger
 import traceback
 from queue import Queue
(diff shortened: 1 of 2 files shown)


kezhenxu94 (approved):
LGTM, thanks

PR #91, opened 2020-11-25
Add sanic plugin

fuhuo:
- [x] Add a test case for the new plugin

### DIFF
--- a/skywalking/plugins/sw_sanic.py
+++ b/skywalking/plugins/sw_sanic.py
@@ -30,8 +30,6 @@ version_rule = {
     "rules": [">=20.3.0"]
 }
 
-_status_code = 0
-
 
 def install():
     from inspect import isawaitable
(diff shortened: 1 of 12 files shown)


kezhenxu94:
@fuhuo thanks for contributing, can you please fix the CI and check the item list in the?

fuhuo:
ok, i'll try

kezhenxu94:
> ok, i'll try

If you need any help, feel free to ping me

kezhenxu94:
Please check this one by one before we can merge this PR, thanks, if you need help, feel free to ping me anytime

image

skywalking/plugins/sw_sanic.py
```diff
@@ -0,0 +1,89 @@
...
+    from sanic import Sanic, handlers, response
+
+    _format_http1_response = response.format_http1_response
+    _handle_request = Sanic.handle_request
+    _handlers_ErrorHandler_reponse = handlers.ErrorHandler.response
+
+    def _sw_format_http1_reponse(status: int, headers, body=b""):
+        global _status_code
+        _status_code = status
+        return _format_http1_response(status, headers, body)
```
kezhenxu94, line 47: Please don't use a global variable to store the status code, in concurrent scenario, the status code may mess up in different requests.

You can use the same way as line 81 to get the current active span (nullablity check) and set the status code here

```python
entry_span = get_context().active_span()
```

kezhenxu94 (commented):
Also, could you please also update the plugin list [here](https://github.com/apache/skywalking-python/blob/master/docs/Plugins.md), thanks very much

skywalking/plugins/sw_sanic.py
fuhuo: ok,thanks.i'll try to fix it later

kezhenxu94 (approved):
Thanks very much for your contributions and patience

PR #95, opened 2020-12-05
test: make tests really run on specified Py version and deprecate 3.5 as it's EOL

kezhenxu94:
[Python 3.5 EOL](https://www.python.org/downloads/release/python-3510/)

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -30,6 +30,8 @@ jobs:
       matrix:
         python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
       fail-fast: false
+    env:
+      SW_PYTHON_VERSION: ${{ matrix.python-version }}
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -40,10 +42,22 @@ jobs:
...
+        if: ${{ matrix.python-version != '3.5' }}
         run: make setup install
       - name: Lint codes
+        if: ${{ matrix.python-version != '3.5' }}
         run: make lint
       - name: Check license header
+        if: ${{ matrix.python-version != '3.5' }}
         run: make license
       - name: Run unit tests
+        if: ${{ matrix.python-version != '3.5' }}
```
kezhenxu94, line 54: We need this to make it pass the check status first

fgksgf (approved):

PR #96, opened 2020-12-09
[Enhancement] optimized path trace ignore

tom-pytel:
* Removed `SW_TRACE_IGNORE` and just using `SW_TRACE_IGNORE_PATH` to enable path trace ignore.
* Optimized path ignore by merging path and suffix into a single regular expression test.
* decorator.py exception wrapping not necessary because exceptions are checked in `span.__exit__()` on context exit.


kezhenxu94 (approved):
LGTM

PR #98, opened 2020-12-10
[Bugfix] allow pending data to send before exit

tom-pytel:
Fixed the problem of last pending data not being sent on exit.

### DIFF
--- a/skywalking/agent/__init__.py
+++ b/skywalking/agent/__init__.py
@@ -15,12 +15,13 @@
 # limitations under the License.
 #
 
-from skywalking.loggings import logger
+import atexit
 from queue import Queue
 from threading import Thread, Event
 from typing import TYPE_CHECKING
 
-from skywalking import config, plugins
+from skywalking import config, plugins, loggings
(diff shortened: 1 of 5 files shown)


kezhenxu94 (dismissed):
LGTM

skywalking/agent/__init__.py
```diff
@@ -38,7 +39,8 @@ def __heartbeat():
 def __report():
     while not __finished.is_set():
         if connected():
-            __protocol.report(__queue)  # is blocking actually
+            while __protocol.report(__queue):  # blocking but has timeout
+                pass
```
kezhenxu94, line 43: So maybe this is unnecessary after your latest commit? And there is no timeout actually (the comment says it has, which is confusing)

tom-pytel: You are right on both counts, the `return True` from `Protocol.report()` also becomes unnecessary, stuck around from first commit.

kezhenxu94 (approved):

PR #99, opened 2020-12-11
Fix: queue.get(block=False) can raise Empty

tom-pytel:
Minor fix for previous PR, calling `queue.get(block=False)` can raise `_queue.Empty` which would print out an exception on exit.


kezhenxu94 (approved):
LGTM

PR #100, opened 2020-12-11
[WIP] started on aiohttp

tom-pytel:
Started on new plugin for aiohttp, requester works but still need to add server component. Again hoping you can handle the tests for this. Will see about adding component ID and maybe logo myself when it is done.

- [ ] Add a component id in [the main repo](https://github.com/apache/skywalking/blob/master/oap-server/server-bootstrap/src/main/resources/component-libraries.yml#L415)
- [ ] Add a logo in [the UI repo](https://github.com/apache/skywalking-rocketbot-ui/tree/master/src/views/components/topology/assets)
- [ ] Rebuild the `requirements.txt` by running `tools/env/build_requirements_(linux|windows).sh`

### DIFF
--- a/skywalking/agent/protocol/grpc.py
+++ b/skywalking/agent/protocol/grpc.py
@@ -18,7 +18,7 @@
 import logging
 from skywalking.loggings import logger
 import traceback
-from queue import Queue
+from queue import Queue, Empty
 
 import grpc
 
@@ -70,7 +70,10 @@ class GrpcProtocol(Protocol):
     def report(self, queue: Queue, block: bool = True):
         def generator():
             while True:
-                segment = queue.get(block=block)  # type: Segment
+                try:
+                    segment = queue.get(block=block)  # type: Segment
+                except Empty:
+                    return
 
                 logger.debug('reporting segment %s', segment)
 
--- a/skywalking/agent/protocol/http.py
+++ b/skywalking/agent/protocol/http.py
@@ -16,7 +16,7 @@
 #
 
 from skywalking.loggings import logger
-from queue import Queue
+from queue import Queue, Empty
 
 from skywalking.agent import Protocol
 from skywalking.client.http import HttpServiceManagementClient, HttpTraceSegmentReportService
@@ -41,7 +41,10 @@ class HttpProtocol(Protocol):
     def report(self, queue: Queue, block: bool = True):
         def generator():
             while True:
-                segment = queue.get(block=block)  # type: Segment
+                try:
+                    segment = queue.get(block=block)  # type: Segment
+                except Empty:
+                    return
 
                 logger.debug('reporting segment %s', segment)
 
--- a/skywalking/agent/protocol/kafka.py
+++ b/skywalking/agent/protocol/kafka.py
@@ -17,7 +17,7 @@
 
 import logging
 from skywalking.loggings import logger, getLogger
-from queue import Queue
+from queue import Queue, Empty
 
 from skywalking import config
 from skywalking.agent import Protocol
@@ -45,7 +45,10 @@ class KafkaProtocol(Protocol):
     def report(self, queue: Queue, block: bool = True):
         def generator():
             while True:
-                segment = queue.get(block=block)  # type: Segment
+                try:
+                    segment = queue.get(block=block)  # type: Segment
+                except Empty:
+                    return
 
                 logger.debug('reporting segment %s', segment)
 


tom-pytel:
Here is an async test snipped:
```py
import asyncio
import aiohttp

from skywalking import agent
from skywalking.decorators import trace
from skywalking.trace.context import get_context
import skywalking.trace.context as context

cities = ['paris', 'lyon', 'nice']

agent.start()

async def child(i):
    async with aiohttp.ClientSession() as session:
        url = f"https://geo.api.gouv.fr/communes?nom={cities[i]}&fields=nom,region&format=json&geometry=centr"

        async with session.get(url) as response:
            return await response.json()

loop = asyncio.get_event_loop()

with get_context().new_local_span('/parent'):
    loop.run_until_complete(
        asyncio.gather(
            *(trace(f'child{i}')(child)(i) for i in range(3))
        )
    )
```

skywalking/plugins/sw_aiohttp.py
```diff
@@ -0,0 +1,65 @@
...
+            ((url.user or '') + ':' + (url.password or '') + '@' if url.user or url.password else '') + \
+            (url.host or '') + \
+            (':' + str(url.explicit_port) if url.explicit_port is not None else '')
+        context = get_context()
+
+        with context.new_exit_span(op=url.path or "/", peer=peer) as span:
+            span.layer = Layer.Http
+            span.component = Component.Unknown  # TODO: add Component.aiohttp
+            span.tag(Tag(key=tags.HttpMethod, val=method.upper()))
+            span.tag(Tag(key=tags.HttpUrl, val=url))
```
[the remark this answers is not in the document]
TomMD: Congratulations on the 100th pull request for skywalking-python!

I think this is the second "missing argument" error on Tag.__init__.  If you'd like I can make a PR (or you are free to) that adds `igore = [ "Missing argument" ]` to the config.toml.  Otherwise I'm likely to do something in the backend that removes this seemingly common seemingly false positive.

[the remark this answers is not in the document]
tom-pytel: The weird thing is the same code doesn't trigger warning in the other plugins.

skywalking/plugins/sw_aiohttp.py
```diff
@@ -0,0 +1,65 @@
...
+def install():
+    from aiohttp import ClientSession
+    from multidict import CIMultiDict, MultiDict, MultiDictProxy
+    from yarl import URL
+
+    async def _sw_request(self: ClientSession, method: str, str_or_url, **kwargs):
+        url = URL(str_or_url)
+        peer = \
+            (url.scheme + '://' if url.scheme else '') + \
+            ((url.user or '') + ':' + (url.password or '') + '@' if url.user or url.password else '') + \
```
kezhenxu94, line 33: I don’t think we want to include `password` here

skywalking/plugins/sw_aiohttp.py
```diff
@@ -0,0 +1,65 @@
...
+            span.tag(Tag(key=tags.HttpMethod, val=method.upper()))  # pyre-ignore
+            span.tag(Tag(key=tags.HttpUrl, val=url))  # pyre-ignore
+
+            carrier = span.inject()
+            headers = kwargs.get('headers')
+
+            if headers is None:
+                headers = CIMultiDict()
+            elif not isinstance(headers, (MultiDictProxy, MultiDict)):
+                headers = CIMultiDict(headers)
```
kezhenxu94, line 50: Should put `headers` back to `kwargs`? Also, is it safe to "override" the existing headers if it's not `MultiDic(Proxy)`?

skywalking/plugins/sw_aiohttp.py
tom-pytel: Yes its probably better not to. The thing is I copied this behavior from the other plugins so they will need to be changed to explicitly remove the password as well since `urllib.parse` always returns it as part of `netloc` if it is present and so would be recorded. Then the Node agent also, as well I am wondering if the Java plugins record this?

skywalking/plugins/sw_aiohttp.py
tom-pytel: Put `headers` back to `kwargs`, yes I meant to do this but forgot, this is why reviews are good :smiley: 

As for "overriding", as far as I can see it should be ok since `aiohttp` doesn't do anything with the headers until it executes the following in `ClientSession._prepare_headers()`:
```py
if not isinstance(headers, (MultiDictProxy, MultiDict)):
    headers = CIMultiDict(headers)
```
Which is actually where I got it from, and yes `CIMultiDict` is an instance of `MultiDict` so I am just executing what would be done anyway a little earlier.

skywalking/plugins/sw_aiohttp.py
kezhenxu94: > Yes its probably better not to. The thing is I copied this behavior from the other plugins so they will need to be changed to explicitly remove the password as well since `urllib.parse` always returns it as part of `netloc` if it is present and so would be recorded. Then the Node agent also, as well I am wondering if the Java plugins record this?

The Java agent doesn't record this

tom-pytel: > The Java agent doesn't record this

So just Python and Node, both incoming and outgoing, http and any other protocols (layers)?

kezhenxu94: > > The Java agent doesn't record this
> 
> 
> 
> So just Python and Node, both incoming and outgoing, http and any other protocols (layers)?

Do you mean peer or password? We usually only record host/ip and port in `peer`

tom-pytel:
Test snippet:
```py
from aiohttp import web
from skywalking import agent

agent.start()

async def handle(request):
    name = request.match_info.get('name', "Anonymous")
    text = "Hello, " + name
    return web.Response(text=text)

app = web.Application()
app.add_routes([web.get('/', handle),
                web.get('/{name}', handle)])

if __name__ == '__main__':
    web.run_app(app, port=8000)
```
Am I missing anything to be stored or handled?

skywalking/plugins/sw_aiohttp.py
tom-pytel: Like for example in sw_requests.py, the following code will store the user and password if I am not mistaken:
```py
    def _sw_request(this: Session, method, url,
                    params=None, data=None, headers=None, cookies=None, files=None,
                    auth=None, timeout=None, allow_redirects=True, proxies=None,
                    hooks=None, stream=None, verify=None, cert=None, json=None):

        from urllib.parse import urlparse
        url_param = urlparse(url)

        ...

        context = get_context()
        with context.new_exit_span(op=url_param.path or "/", peer=url_param.netloc) as span:
```
I thinkthis is repeated across several plugins.

kezhenxu94: We need to review these plugins and if they really did store username/password, we need to remove them explicitly, usually as the name `peer` indicates, we only store host/IP and port into it

kezhenxu94:
> Test snippet:
> 
> ```python
> from aiohttp import web
> from skywalking import agent
> 
> agent.start()
> 
> async def handle(request):
>     name = request.match_info.get('name', "Anonymous")
>     text = "Hello, " + name
>     return web.Response(text=text)
> 
> app = web.Application()
> app.add_routes([web.get('/', handle),
>                 web.get('/{name}', handle)])
> 
> if __name__ == '__main__':
>     web.run_app(app, port=8000)
> ```
> 
> Am I missing anything to be stored or handled?

I’ll test locally, also, can you please uncomment the PR template so that we can check the items one by one (some may be checked by me though)

kezhenxu94:
Hi @tom-pytel , sorry for closing this by mistake because of a force push, can you please reopen another PR?

PR #105, opened 2020-12-17
Fix Reconnect when report has an exception occurs

wuchengyi2006:

### DIFF
--- a/skywalking/agent/__init__.py
+++ b/skywalking/agent/__init__.py
@@ -39,9 +39,19 @@ def __heartbeat():
 def __report():
     while not __finished.is_set():
         if connected():
-            __protocol.report(__queue)  # is blocking actually
-
-        __finished.wait(1)
+            try:
+                __protocol.report(__queue)
+            except Exception as e:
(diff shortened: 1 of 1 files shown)


kezhenxu94:
Hey @wuchengyi2006 , thanks! The exception should be caught here https://github.com/apache/skywalking-python/blob/37edeba1d4cab668e9ab21fd00b3be1b5ec8c18d/skywalking/agent/protocol/grpc.py#L121-L124

which version are you using and what exception did you encounter that is not caught by 👆

wuchengyi2006:
> Hey @wuchengyi2006 , thanks! The exception should be caught here
> 
> https://github.com/apache/skywalking-python/blob/37edeba1d4cab668e9ab21fd00b3be1b5ec8c18d/skywalking/agent/protocol/grpc.py#L121-L124
> 
> which version are you using and what exception did you encounter that is not caught by 👆

Great,My version maybe is too old,It seems that the problem has been solved！

kezhenxu94:
Thanks for confirmation

PR #107, opened 2020-12-24
Add changelog for 0.5.0

kezhenxu94:

### DIFF
--- a/.asf.yaml
+++ b/.asf.yaml
@@ -35,6 +35,4 @@ github:
         strict: true
         contexts:
           - CheckStatus
-      required_pull_request_reviews:
-        dismiss_stale_reviews: true
-        required_approving_review_count: 1
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,29 @@
 ## Change Logs
 
+### 0.5.0
+
+- New plugins
+    - Pyramid Plugin (#102)
+    - AioHttp Plugin (#101)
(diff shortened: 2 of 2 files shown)


.asf.yaml
```diff
@@ -35,6 +35,4 @@ github:
         strict: true
         contexts:
           - CheckStatus
-      required_pull_request_reviews:
-        dismiss_stale_reviews: true
-        required_approving_review_count: 1
```
kezhenxu94, line 40: I'm removing this since trivial changes like this PR can be merged quickly as long as the checks passed

fgksgf: How about non-trivial changes ?

kezhenxu94: > How about non-trivial changes ?

Committers can request reviews from other committers, as we have limited committers in this repo, I'd rather work on trust that we know the changes are non-trivial and need review

fgksgf: Can those people who are not members (committer/PMC) merge directly when the checks passed ? If they can't I think this change of `.asf.yml` is ok.

kezhenxu94: > Can those people who are not members (committer/PMC) merge directly when the checks passed ? If they can't I think this change of .asf.yml is ok.

Of course they can't, merging needs write access to the repo

fgksgf: Ok, I have no worries.

fgksgf (approved):

PR #110, opened 2020-12-24
Set up license-eye to check license

kezhenxu94:

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -37,6 +37,10 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: true
+      - name: Check License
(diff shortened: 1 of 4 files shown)


.github/workflows/build.yaml
```diff
@@ -37,6 +37,10 @@ jobs:
         uses: actions/checkout@v2
         with:
           submodules: true
+      - name: Check License
+        uses: apache/skywalking-eyes@main
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
wu-sheng, line 43: Should we officially merge this after the eye 0.1.0 release?

kezhenxu94: Yes, this can be done after license-eye is released, but now just to verify it works

PR #114, opened 2021-01-05
Sync the meter protocol to the Python agent

kezhenxu94:
This patch synchronizes the meter protocol so that the contributors can use it to implement the meter report feature


wu-sheng (approved):
Is there potential one already? This sync seems like a preparation.

kezhenxu94:
> Is there potential one already? This sync seems like a preparation.

Yes, we have potential contributor interested in this

PR #122, opened 2021-06-05
Initial psycopg plugin

tom-pytel:
I will do the other plugin stuff once the test is stable.

### DIFF
--- a/skywalking/__init__.py
+++ b/skywalking/__init__.py
@@ -31,6 +31,7 @@ class Component(Enum):
     Tornado = 7005
     Redis = 7
     MongoDB = 9
+    PostgreSQL = 22
     KafkaProducer = 40
     KafkaConsumer = 41
     RabbitmqProducer = 52
@@ -40,7 +41,6 @@ class Component(Enum):
     Sanic = 7007
     AioHttp = 7008
     Pyramid = 7009
-    Psycopg = 7010
 
 
 class Layer(Enum):
--- a/skywalking/__init__.py
+++ b/skywalking/__init__.py
@@ -40,6 +40,7 @@ class Component(Enum):
     Sanic = 7007
     AioHttp = 7008
     Pyramid = 7009
+    Psycopg = 7010
 
 
 class Layer(Enum):
--- a/tests/plugin/sw_psycopg2/docker-compose.yml
+++ b/tests/plugin/sw_psycopg2/docker-compose.yml
@@ -1,0 +1,79 @@
+#
+# Licensed to the Apache Software Foundation (ASF) under one or more
+# contributor license agreements.  See the NOTICE file distributed with
+# this work for additional information regarding copyright ownership.
+# The ASF licenses this file to You under the Apache License, Version 2.0
+# (the "License"); you may not use this file except in compliance with
+# the License.  You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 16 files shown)


wu-sheng:
@tom-pytel Take a look at slack. I left a message for your Apache committer to further process. It is in your hands to complete them.

tom-pytel:
Hey @kezhenxu94, I have no idea why the test is failing. I copied the framework from pymsql test which works and the code in `provider.py` works by itself but for some reason running as a test it is getting:
```
Exception: Wait time exceeded 100 sec. Exception HTTPConnectionPool(host='0.0.0.0', port=9090): Max retries exceeded with url: /users (Caused by NewConnectionError('<urllib3.connection.HTTPConnection object at 0x7fc586e6b040>: Failed to establish a new connection: [Errno 111] Connection refused'))
```

tests/plugin/sw_psycopg2/docker-compose.yml
```diff
@@ -0,0 +1,79 @@
...
+    image: postgres:11
+    hostname: postgres
+    ports:
+      - 5432:5432
+    environment:
+      - POSTGRES_USER=root
+      - POSTGRES_PASSWORD=root
+      - POSTGRES_DB=admin
+    healthcheck:
+      test: ["CMD", "bash", "-c", "cat < /dev/null > /dev/tcp/127.0.0.1/3306"]
```
kezhenxu94, line 36: Hi, this is the reason why tests failed, the exposed port is 5432 but this checks 3306, I believe it's a copy-paste typo.

Screen Shot 2021-06-06 at 09 42 31

```suggestion
      test: ["CMD", "bash", "-c", "cat < /dev/null > /dev/tcp/127.0.0.1/5432"]
```

tom-pytel:
Ok, one thing to resolve before merge. I reused `config.mysql_trace_sql_parameters` and `config.mysql_sql_parameters_max_length` options in this plugin just to get it running but the final param should be one of two cases:
1. Add separate `config.postgre_trace_sql_parameters` and `config.postgre_sql_parameters_max_length`.
2. Convert `config.mysql_trace_sql_parameters` to just `config.trace_sql_parameters` and have all SQL agents share this to avoid explosion of options as different SQL agents are added (since odds are you will not be using mysql and postgresql simultaneously in the same project) - this is my preferred option.

And another idea unrelated to which option above is taken. If you are not locked into the current options scheme - reduce the two parameters `config.mysql_trace_sql_parameters` and `config.mysql_sql_parameters_max_length` into a single option `config.mysql_parameters` (or something like that) which gives the length and if length 0 is specified (default) then disables parameters?

kezhenxu94:
> And another idea unrelated to which option above is taken. If you are not locked into the current options scheme - reduce the two parameters `config.mysql_trace_sql_parameters` and `config.mysql_sql_parameters_max_length` into a single option `config.mysql_parameters` (or something like that) which gives the length and if length 0 is specified (default) then disables parameters?

This is an optimization I've been thinking about, I'm +1 to it.

Also, I prefer (2.), since there is rare situation that a service will use multiple DB layers and not use the same configurations for them.

skywalking/__init__.py
```diff
@@ -40,6 +40,7 @@ class Component(Enum):
     Sanic = 7007
     AioHttp = 7008
     Pyramid = 7009
+    Psycopg = 7010
```
kezhenxu94, line 43: Remember to add this in main repo and UI repo

kezhenxu94 (commented):
Please also add the plugin to https://github.com/apache/skywalking-python/blob/master/docs/Plugins.md 🙇

skywalking/__init__.py
tom-pytel: I've got that done just was waiting for this part to finalize.

kezhenxu94 (approved):

PR #126, opened 2021-06-28
Support agent namespace.

wuwen5:
Support the agent namespace, keep consistent with java side


kezhenxu94 (approved):
Thank you @wuwen5 very much, 🙇🏻

PR #127, opened 2021-07-11
Query and dispatch ProfileTaskCommand

Humbertzhang:
Add query and dispatch part of ProfileTaskCommand, mainly includes those steps:

```
1. query ProfileTaskCommand from OAPServer.
2. deserialize command to object
3. dispatch it to appropriate CommandExecutor, here is ProfileTaskCommandExecutor
4. check the ProfileTask's attributes is rational and add it in profile task queue
```
It uses two threads to complete the task: `QueryProfileCommandThread` for query and `CommandDispatchThread` for command dispatch and check.

This PR is the first part of https://github.com/apache/skywalking/issues/5943. I will add the execute part and tests for profiling feature in following PRs.

PR #128, opened 2021-07-11
removed requests as required dependency

tom-pytel:
Fix https://github.com/apache/skywalking/issues/7282 due to #125

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -39,7 +39,6 @@ setup(
         "grpcio",
         "grpcio-tools",
         "packaging",
-        "requests",
         "wrapt",
     ],
     extras_require={


tom-pytel:
So if you have nothing against it I will look into replacing requests with standard python lib urllib to do the connections?

wu-sheng (approved):
Do we need

wu-sheng:
FYI @potiuk This has been reverted and kept optional only.

tom-pytel:
> Do we need

A different communication method? If requests is license problematic then why not change to standard python lib if possible, one less dependency too.

wu-sheng:
> > Do we need
> 
> A different communication method? If requests is license problematic then why not change to standard python lib if possible, one less dependency too. https://github.com/psf/requests/pull/5797. We just need their new release.

Sorry, it was a copy/paste issue. There is no issue. This dependency issue should be resolved from the upstream repo already.

tom-pytel:
Heads up, I just grabbed the latest master which should have this and #127 merged and this change is not present. `setup.py` still contains `requests` as a required dependency. Guessing merge with @127 list it?

kezhenxu94:
> Heads up, I just grabbed the latest master which should have this and #127 merged and this change is not present. `setup.py` still contains `requests` as a required dependency. Guessing merge with @127 list it?

Hey, I can see `requests` is not required dependency. https://github.com/apache/skywalking-python/blob/master/setup.py It’s in `http` extra dependency.

potiuk:
Correct. I mentioned that in my report. So this is right, it's not *necesssary* according to the ASF Policy (similarly in Airflow we have a lot of extras with `requests`) - the "optional" LGPL dependency is acceptable. But once the new version is out, I think it would be great to make it "minimal" version of requests - this way the LGPL dependency becomes optional even for optional dependencies ;)

PR #130, opened 2021-07-15
removed trace/ipc

tom-pytel:
This explicit method no longer needed since `fork()` handling is now automatic via `os.register_at_fork()`.


tom-pytel:
Added `grpc.max_connection_age_grace_ms = 1000 * config.GRPC_TIMEOUT` to `grpc.insecure_channel()`. According to the documentation this is supposed to default to 30 minutes. In practice if this is not explicitly set here with current value of 5 minutes then timeouts happen all the time even if spans are being constantly sent. Strange...

kezhenxu94 (approved):

PR #135, opened 2021-07-24
update init param name

zgfh:
update init param name


kezhenxu94 (commented):
For master branch, we recently change the init param names, if you are using previous version, please read the corresponding doc

zgfh:
thinks,good to know it

PR #136, opened 2021-07-24
fix trace decorator not work

zgfh:
fix trace decorator not work


wu-sheng:
Please explain why it is not working, and provide the cases.

tom-pytel:
See #137, let me know if fixed.

zgfh:
@tom-pytel it work for me ,thanks

PR #149, opened 2021-08-10
Enable HTTP log reporting

Superskyyy:
This is a work in progress. 

Now reporting logs in JSON through HTTP to `http://oap/v3/logs` does work. 

But, I'm not sure if the `oap/v3/logs` endpoint is for such usage(It seems designed for fluent-bit batch reporting?). Reporting logs one by one through HTTP may not be ideal in terms of performance. I'm not sure whether the Java agent only implements gRPC reporter intentionally out of this reason. 

Please advise.

  
Signed-off-by: YihaoChen <Superskyyy@outlook.com>

### DIFF
--- a/skywalking/client/http.py
+++ b/skywalking/client/http.py
@@ -123,7 +123,7 @@ class HttpLogDataReportService(TraceSegmentReportService):
         self.session = requests.Session()
 
     def report(self, generator):
-        for log_data in generator:
-            json_string = json_format.MessageToJson(log_data)
-            res = self.session.post(self.url_report, json=[json.loads(json_string)])
-            logger.debug('report log response: %s', res)
+        log_batch = [json.loads(json_format.MessageToJson(log_data)) for log_data in generator]
+        if log_batch:  # prevent empty batches
+            res = self.session.post(self.url_report, json=log_batch)
+            logger.debug('report batch log response: %s', res)
--- a/skywalking/client/http.py
+++ b/skywalking/client/http.py
@@ -14,13 +14,14 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 #
-
-from skywalking.loggings import logger
+import json
 
 import requests
+from google.protobuf import json_format
 
 from skywalking import config
 from skywalking.client import ServiceManagementClient, TraceSegmentReportService
+from skywalking.loggings import logger
 
 
 class HttpServiceManagementClient(ServiceManagementClient):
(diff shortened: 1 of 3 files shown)


tom-pytel:
> But, I'm not sure if the `oap/v3/logs` endpoint is for such usage(It seems designed for fluent-bit batch reporting?). Reporting logs one by one through HTTP may not be ideal in terms of performance. I'm not sure whether the Java agent only implements gRPC reporter intentionally out of this reason.

Not sure what the INTENDED usage of that endpoint was, but the same individual send inefficiency applies to how the spans are sent currently to `/v3/segment` instead of batching to `/v3/segments`, this should be optimized at some point. The hit is not that bad though since the http protocol uses `requests.Session` which should do persistent connections as per https://docs.python-requests.org/en/master/user/advanced/:

"The Session object allows you to persist certain parameters across requests. It also persists cookies across all requests made from the Session instance, **and will use urllib3’s connection pooling. So if you’re making several requests to the same host, the underlying TCP connection will be reused**, which can result in a significant performance increase (see HTTP persistent connection)."

In any case, if it works but may not yet be optional it is a step forward.

skywalking/client/http.py
```diff
@@ -109,3 +110,20 @@ def report(self, generator):
...
+        self.session = requests.Session()
+
+    def fork_after_in_child(self):
+        self.session.close()
+        self.session = requests.Session()
+
+    def report(self, generator):
+        for log_data in generator:
+            json_string = json_format.MessageToJson(log_data)
+            res = self.session.post(self.url_report, json=[json.loads(json_string)])
```
tom-pytel, line 128: Actually, looking at this suggests that the `/v3/logs` endpoint can take an array of logs so a batch send could be done. Change to:
```py
def report(self, generator):
    json = [json_format.MessageToJson(log_data) for log_data in generator]
    res = self.session.post(self.url_report, json=json)
```
Your call if you want to do now.

tom-pytel (approved):

skywalking/client/http.py
Superskyyy: Umm, should I make a new config entry allowing the user to choose whether batch or not?

tom-pytel: No, batch is the right way to go (assuming this endpoint does take arrays, which is what you should check).

Superskyyy: Got it, it does take arrays.

Humbertzhang:
So far looks good to me, looking forward to your Kafka part, thank you.

tom-pytel:
I would say merge this first then do Kafka as a separate PR.

kezhenxu94:
> I would say merge this first then do Kafka as a separate PR.

I agree. @Superskyyy let's do one thing at a time, in a single PR

Superskyyy:
> > I would say merge this first then do Kafka as a separate PR.
> 
> I agree. @Superskyyy let's do one thing at a time, in a single PR

No problem, but let me add a commit on the batch reporting first. Then lets merge.

Superskyyy:
Oops, messed up a bit.

kezhenxu94:
> Oops, messed up a bit.

It's ok, we will squash the commits into one when merging

kezhenxu94:
@Superskyyy ping me when it's ready to merge

Superskyyy:
@kezhenxu94 Checks done, ready to merge.

kezhenxu94:
> @kezhenxu94 Checks done, ready to merge.

Thank you @Superskyyy very much 🙇🏻 , excellent work!

PR #150, opened 2021-08-11
[plugin] pymongo compatible with mongodb cluster

cooolr:
replace the `this.collection.database.client.address` with the `this.collection.database.client.nodes` to be compatible with the mongodb cluster

### DIFF
--- a/skywalking/plugins/sw_pymongo.py
+++ b/skywalking/plugins/sw_pymongo.py
@@ -101,8 +101,8 @@ def inject_bulk_write(_Bulk, bulk_op_map):
     _execute = _Bulk.execute
 
     def _sw_execute(this: _Bulk, *args, **kwargs):
-        address = this.collection.database.client.address
-        peer = "%s:%s" % address
+        nodes = this.collection.database.client.nodes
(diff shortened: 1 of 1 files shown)


skywalking/plugins/sw_pymongo.py
```diff
@@ -101,8 +101,8 @@ def inject_bulk_write(_Bulk, bulk_op_map):
     _execute = _Bulk.execute
 
     def _sw_execute(this: _Bulk, *args, **kwargs):
-        address = this.collection.database.client.address
-        peer = "%s:%s" % address
+        nodes = this.collection.database.client.nodes
```
kezhenxu94, line 104: So when the MongoDB is in single-node mode, the `this.collection.database.client.nodes` will also contain one element?

kezhenxu94 (approved):
@cooolr thank you!

Humbertzhang (approved):
LGTM

skywalking/plugins/sw_pymongo.py
Humbertzhang: I looked at it and the answer is yes.

PR #151, opened 2021-08-11
Fixes and tweaks to recent falcon plugin

tom-pytel:
Some tweaks to #146

* Operation endpoint name gotten from url path instead of being whole URL.
* Added ignore by http method.
* Port of peer was incorrectly set from server port.
* Headers gotten directly from request object instead of parsing environment.
* Order of parameters in _handle_exception was wrong.
* HTTP parameters joined by comma instead of newline.
* The line "resp_obj = response.Response(ResponseOptions())" was just creating a new clean response without taking into account any errors from the current request, so errors like 404 were not marked as errors and returned with code 200.
* Replaced _sw_handle_exception with a direct wrap of call to handler just in case registered exception handlers suppress the exception.
* Added TagHttpStatusMsg.

### DIFF
--- a/skywalking/plugins/sw_falcon.py
+++ b/skywalking/plugins/sw_falcon.py
@@ -15,84 +15,65 @@
 # limitations under the License.
 #
 
-from skywalking import Layer, Component
+from skywalking import Layer, Component, config
 from skywalking.trace.carrier import Carrier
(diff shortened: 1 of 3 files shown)


tom-pytel:
This is good to go.

kezhenxu94 (approved):

PR #152, opened 2021-08-11
Remove HTTP basic auth credentials from log, stacktrace, segment

Superskyyy:
- Removes HTTP basic auth `user:pw` from span - `peer` `url` and `log` in http client libs
- Removes HTTP basic auth `user:pw` from log and tracebacks in log reporter
- Adds support for an extra configuration - `cause_exception_depth` which limits the stacktrace depth (align with Java agent)
- Updates some doc and config naming.

- Tested working for (requests, urllib3, urllib.request, aiohttp) under various settings, only two needs span fix.

Fixes https://github.com/apache/skywalking/issues/6018

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,15 @@
 
 ### 0.7.0
 
-- New plugins
+- Feature:
+    - Support collecting and reporting logs to backend (#147)
+
+- New plugins:
     - Falcon Plugin (#146)
 
+- Fixes:
+    - Now properly removes HTTP basic auth credentials from segments and logs (#152)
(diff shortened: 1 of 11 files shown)


kezhenxu94:
Hi @Superskyyy , I saw you only update 2 http plugins to use the filters, did you check whether or not other http plugins need the same?

Superskyyy:
> Hi @Superskyyy , I saw you only update 2 http plugins to use the filters, did you check whether or not other http plugins need the same?

I checked the other two are fine, they seem to avoided that already. Also tested with some sample code.

PR #156, opened 2021-08-23
Support loading Python agent through sw-python CLI

Superskyyy:
This PR adds functionality of non-intrusive integration CLI for the Python agent.

Though the code is completed, as a dangerous feature that modifies Python bootstrap I will need to test it a bit more before it's ready for review.

closes https://github.com/apache/skywalking/issues/7381


kezhenxu94:
Looking at the docs just makes me feel exciting about this new feature that I've been planning for so long, and I believe this feature will be the last piece before our 1.0 GA. Look forward to this feature and our 1.0 GA version, soon!

kezhenxu94:
Let me review and try this new CLI tomorrow 🤩

Superskyyy:
> Looking at the docs just makes me feel exciting about this new feature that I've been planning for so long, and I believe this feature will be the last piece before our 1.0 GA. Look forward to this feature and our 1.0 GA version, soon!

That's what I thought! This feature will ease a lot of work in deployment.

> Let me review and try this new CLI tomorrow 🤩

I also need to run it more in a Linux env tmr, I'm struggling with my WSL and Windows doesn't communicate to each other properly over localhost... So my local test coverage is quite limited for now (not able to deploy gunicorn on windows etc.). 

After I verify most cases, I could use some help testing this CLI on different OS and shells and see if it breaks for some corner cases.

wu-sheng:
Use GHA to verify running in various env.

cli.py
```diff
@@ -0,0 +1,26 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+""" Just an entry point script
+python -m cli run command
```
kezhenxu94, line 19: Is this name `cli` (and the file name `cli.py`) too generous? What about using `sw-python` to make it consistent with our standalone command?

docs/CLI.md
```diff
@@ -0,0 +1,87 @@
+# SkyWalking Python Agent Command-Line Interface(CLI)
+
+You would at least need to add the following lines to your applications in previous releases to get the agent attached and running.
```
kezhenxu94, line 3: What about putting the condition at the beginning? In case the users just read a half of the sentence 🤣 and let's be clear about since which version we support this

```suggestion
In earlier releases than 0.7.0, you would at least need to add the following lines to your applications to get the agent attached and running.  
```

docs/CLI.md
```diff
@@ -0,0 +1,87 @@
...
+```
+
+
+Now the SkyWalking Python agent implements a command-line interface that can be utilized to attach the agent to your
+awesome applications during deployment **without changing any application code**, 
+just like the [SkyWalking Java Agent](https://github.com/apache/skywalking).
+
+## Usage
+
+Upon successful installation of the SkyWalking Python agent via pip,
```
kezhenxu94, line 17:
```suggestion
Upon successful [installation of the SkyWalking Python agent via pip](../README.md#install),
```

docs/CLI.md
```diff
@@ -0,0 +1,87 @@
...
+Now the SkyWalking Python agent implements a command-line interface that can be utilized to attach the agent to your
+awesome applications during deployment **without changing any application code**, 
+just like the [SkyWalking Java Agent](https://github.com/apache/skywalking).
+
+## Usage
+
+Upon successful installation of the SkyWalking Python agent via pip, 
+a command-line script `sw-python` is installed in your environment(virtual env preferred).
+
+### The run option
```
kezhenxu94, line 20:
```suggestion
### The `run` option
```

docs/CLI.md
```diff
@@ -0,0 +1,87 @@
...
+### Configuring the agent 
+
+You would normally want to provide additional configurations other than the default ones.
+
+#### Through environment variables
+
+The currently supported method is to provide the environment variables listed 
+in [EnvVars Doc](EnvVars.md) as instructed in the [README](../README.md).
+
+#### Through a sw-config.yaml
```
kezhenxu94, line 67: One thing we should note when implementing this is that we've been trying to minimize our dependencies as an agent, so we'd better choose a reasonable file format that Python built-in libs can parse, instead of bringing another library, to avoid version conflicts with user applications, `yaml` is a common case where users may already use and we should not require dependency upgrade or downgrade.

skywalking/bootstrap/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+""" This sub-package is for the convenience of deployment and automation
+A CLI for running Python scripts and programs with SkyWalking Python Agent automatically attached.
+`loader/sitecustomize.py` is invoked by the Python interpreter at startup.
+"""
+
+__version__ = '0.6.0'
```
kezhenxu94, line 23: Is there any means we can bump up this version automatically together with the version in `setup.py` when releasing?

kezhenxu94 (commented):
Hi @Superskyyy , I just test (on MacOS) and review the codes, and left some comments.

Since you did a lot of test cases, it would be perfect if you can add those cases into our test matrix in GitHub Actions, if you find any trouble in adding that, ping me at anytime, I can offer my help!

skywalking/bootstrap/loader/sitecustomize.py
```diff
@@ -0,0 +1,140 @@
...
+loaded = sys.modules.pop('sitecustomize', None)  # pop sitecustomize from loaded
+
+# now try to find the original sitecustomize provided in user env
+try:
+    loaded = importlib.import_module('sitecustomie')
+    _sw_loader_logger.debug("Found user sitecustomize file {}, imported".format(loaded))
+except ImportError:  # ModuleNotFoundError
+    _sw_loader_logger.debug("Original sitecustomize module not found, skipping.")
+finally:  # surprise the import error by adding loaded back
+    sys.modules['sitecustomize'] = loaded
```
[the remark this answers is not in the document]
Superskyyy: @sonatype-lift help

Superskyyy:
> Hi @Superskyyy , I just test (on MacOS) and review the codes, and left some comments.
> 
> Since you did a lot of test cases, it would be perfect if you can add those cases into our test matrix in GitHub Actions, if you find any trouble in adding that, ping me at anytime, I can offer my help!

I'm actually thinking what if I change/ duplicate our existing plugin unit tests to use the `sw-python` command in docker-compose command, instead of calling agent in those `consumer` `provider`s, that should work. 

Idk if we need to keep the old tests that directly call `agent.start()`, just that having two versions doing essentially the same thing will double the GHA running time, which seems awfully long.

kezhenxu94:
> I'm actually thinking what if I change/ duplicate our existing plugin unit tests to use the `sw-python` command in docker-compose command, instead of calling agent in those `consumer` `provider`s, that should work.
> 
> Idk if we need to keep the old tests that directly call `agent.start()`, just that having two versions doing essentially the same thing will double the GHA running time, which seems awfully long.

What I think is to replace all existing tests with the `sw-python` as we will recommend users to adapt the later method after it gets merged, we don't need to duplicate the tests.

skywalking/bootstrap/__init__.py
Superskyyy: I think we could move this __version__ to somewhere like the sw_python.py and have both places import that variable. Idk if setup.py can import that during installation, need to try it.

kezhenxu94:
> > I'm actually thinking what if I change/ duplicate our existing plugin unit tests to use the `sw-python` command in docker-compose command, instead of calling agent in those `consumer` `provider`s, that should work.
> > Idk if we need to keep the old tests that directly call `agent.start()`, just that having two versions doing essentially the same thing will double the GHA running time, which seems awfully long.
> 
> What I think is to replace all existing tests with the `sw-python` as we will recommend users to adapt the later method after it gets merged, we don't need to duplicate the tests.

That could be done in following pull requests though.

skywalking/bootstrap/loader/sitecustomize.py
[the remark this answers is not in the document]
Superskyyy: Seems to be a Python typing problem, not solvable.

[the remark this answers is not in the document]
Superskyyy: @sonatype-lift ignore

skywalking/bootstrap/__init__.py
Superskyyy: Nvm that seems a bad idea, I simply removed the version number from CLI output.

kezhenxu94 (approved):
Thank you @Superskyyy very much for doing this!! 🎉

PR #165, opened 2021-09-17
Refactor SkyWalking Python to use the CLI for CI instead of legacy setup

Superskyyy:
Closes https://github.com/apache/skywalking/issues/7701

All tests except Sanic is passing in GHA.

Sanic with `sw-python run` fails for Python 3.6 but not 3.7 -> 3.9, - https://github.com/Superskyyy/skywalking-python/runs/3625969357?check_suite_focus=true 

sanic/provider.py reporting a exit span from 7002(requests) instead of entry span from 7007(sanic), still trying to figure out why.

Please do not merge.


Superskyyy:
kept the original Sanic tests untouched for latest commit, so the CI will pass here.

kezhenxu94:
From the actual data it seems that one of the services (provider, consumer) failed to report data to backend, 2 segments from each respectively are expected while there is only one from one service. Did it consistently fail or did it ever succeed?

might it be the agent failed to send pending data when the application exit?

Superskyyy:
> From the actual data it seems that one of the services (provider, consumer) failed to report data to backend, 2 segments from each respectively are expected while there is only one from one service. Did it consistently fail or did it ever succeed?
> 
> might it be the agent failed to send pending data when the application exit?

It was a consistent failure for CLI based and always succeeds for the original. Just tested more times on my another branch. 

I'm trying to figure out what's the difference I missed between the two testing methods and how it only impacts Python 3.6 zzz.

Superskyyy:
Guess I'll need to make a local Sanic project and bring it up with sw-Python at Python 3.6 for an actual oap to see if anything weird happens.

kezhenxu94:
> Guess I'll need to make a local Sanic project and bring it up with sw-python at Python 3.6 for an actual oap to see if anything weird happens.
> 

For the sake of convenience, consider using the OAP docker image to start an oap instance, and you can just start the Sanic project used in the tests?

Superskyyy:
> > Guess I'll need to make a local Sanic project and bring it up with sw-python at Python 3.6 for an actual oap to see if anything weird happens.
> 
> For the sake of convenience, consider using the OAP docker image to start an oap instance, and you can just start the Sanic project used in the tests?

image

Good news, issue found and fixed(I hope).

The root problem comes from the `from sys import argv` that Sanic uses inside its own code. When the import fails, the `sw_sanic` plugin fails to install, which makes weird errors.

https://bugs.python.org/issue32573

Above mentions the particular limitation in Python 3.6(the bug report claims 3.7 too, but it seems to be working anyway), `sys.argv` is not set when the interpreter is invoked in a way like our CLI, so I applied the workaround to have it always importable.
```Python
if not hasattr(sys, 'argv'):
    sys.argv = ['']

```

kezhenxu94 (approved):
@Superskyyy thank you for taking care of this 🙇🏻

PR #166, opened 2021-09-19
Remove places that mentions Python 3.5 support due to EOL

Superskyyy:
Will test Python 3.10 when officially releases next month.

Signed-off-by: Superskyyy <Superskyyy@outlook.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ microservices, cloud native and container-based (Docker, Kubernetes, Mesos) arch
 
 ## Installation Requirements
 
-SkyWalking Python Agent requires SkyWalking 8.0+ and Python 3.5+.
+SkyWalking Python Agent requires SkyWalking 8.0+ and Python 3.6+.
 
 > If you would like to try out the latest features that are not released yet, please refer to this [guide](docs/en/setup/faq/How-to-build-from-sources.md) to build from sources.
 
--- a/docker/Makefile
+++ b/docker/Makefile
@@ -18,9 +18,8 @@ D := docker
 
 P := grpc http kafka
 
-TARGETS := py3.5 py3.6 py3.7 py3.8 py3.9
+TARGETS := py3.6 py3.7 py3.8 py3.9
 
-py3.5: BASE_IMAGE = python:3.5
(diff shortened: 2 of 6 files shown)


kezhenxu94 (approved):

kezhenxu94:
We can say 3.5 is EOL and is not in our test matrix, it works in versions prior to 1.0, but it may or may not work with the codebase evolves.

Superskyyy:
> We can say 3.5 is EOL and is not in our test matrix, it works in versions prior to 1.0, but it may or may not work with the codebase evolves.

Ok, I intend to swap existing string formatting(`%` `.format` `+` etc.) to f-strings since they are much faster, but that does blocks all Python 3.5 users from v1.0.0, should I do that?

kezhenxu94:
> > We can say 3.5 is EOL and is not in our test matrix, it works in versions prior to 1.0, but it may or may not work with the codebase evolves.
> 
> 
> 
> Ok, I intend to swap existing string formatting(`%` `.format` `+` etc.) to f-strings since they are much faster, but that does blocks all Python 3.5 users from v1.0.0, should I do that?

That's awesome. We can do that, and just say since 1.0 we don't support 3.5.

wu-sheng:
As you said we will drop 3.5 supports, https://twitter.com/ASFSkyWalking/status/1439553219210792963.

README.md
```diff
@@ -19,7 +19,7 @@ microservices, cloud native and container-based (Docker, Kubernetes, Mesos) arch
 
 ## Installation Requirements
 
-SkyWalking Python Agent requires SkyWalking 8.0+ and Python 3.5+.
+SkyWalking Python Agent requires SkyWalking 8.0+ and Python 3.6+.
```
kezhenxu94, line 22: OK, since the requirement here is version-ed, users should check the corresponding doc of their versions, this PR is good to me now

kezhenxu94 (approved):

kezhenxu94:
> Will test Python 3.10 when officially releases next month.
> 

Seems 3.10 is out. @Superskyyy

PR #167, opened 2021-10-19
Cleanup and Python 3.10 support

Superskyyy:
**Changelist:**

1. Evaluates all plugin lib versions.
2. Adds a new utility to support Python version-specific tests (e.g. disable certain tests for certain Python versions).
3. Adds a plugin test table generator utility (in the tools folder, the output is to be copied into the plugins.md doc. for now.)
4. Enhances CI workflow to run more efficiently and new deadlink checker.
5. Supports Python 3.10.
6. Stricter flake8 lint rules.
7. Psycopg3.0 (psycopg) support. - no need to add a component ID.
8. Chore - fix lint, remove unused operationid.

**Note:** 
1. Sanic plugin is problematic, it has incompatible API changes in the next LTS + not supporting Python 310 yet. 
2. Pymongo also has some problems with the next version.
3. Celery is not tested.
4. Psycopg3.0 needs extra instrumentation.

Python 3.6 will EOL on 23 DEC 2021.

### DIFF
--- a/docs/en/setup/Plugins.md
+++ b/docs/en/setup/Plugins.md
@@ -6,7 +6,7 @@ The column of versions only indicates the set of library versions tested in a be
 If you find newer major versions that are missing from the following table, and it's not documented as a limitation, 
 please PR to update the test matrix in the plugin.
 
-Versions marked as NOT SUPPORTED may be due to 
+Versions marked as NOT SUPPORTED may be due to
 an incompatible version with Python in the original library
 or a limitation of SkyWalking auto-instrumentation (welcome to contribute!)
 
@@ -35,6 +35,6 @@ Library | Python Version - Lib Version | Plugin Name
 | [urllib3](https://urllib3.readthedocs.io/en/latest/) | Python >=3.6 - ['1.26', '1.25'];  | `sw_urllib3` |
 | [urllib_request](https://docs.python.org/3/library/urllib.request.html) | Python >=3.6 - ['*'];  | `sw_urllib_request` |
 ### Notes
-- The celery server running with "celery -A ..." should be run with the HTTP protocol 
+- The celery server running with "celery -A ..." should be run with the HTTP protocol
 as it uses multiprocessing by default which is not compatible with the gRPC protocol implementation 
 in SkyWalking currently. Celery clients can use whatever protocol they want.
--- a/skywalking/plugins/sw_aiohttp.py
+++ b/skywalking/plugins/sw_aiohttp.py
@@ -30,6 +30,7 @@ support_matrix = {
 }
 note = """"""
 
+
 def install():
     from aiohttp import ClientSession
     from aiohttp.web_protocol import RequestHandler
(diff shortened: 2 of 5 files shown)


wu-sheng:
> Sanic plugin is problematic, it has incompatible API changes in the next LTS + not supporting Python 310 yet.

Does python agent have version identify capability? Or we have to choose?

tom-pytel:
> **Changelist:**
> 
> 7. Psycopg3.0 (psycopg) support. - no need to add a component ID.

psycopg 3.0 has added an async connection and cursor object which are not instrumented in a simple copy of the psycopg2 plugin. Note should be made of this. I am on a different project now but if I find the time I will see about adding this.

> **Note:**
> 
> 3. Celery is not tested.

My fault, I was short on time. If you want a simple celery script to turn into a full test then I can provide that.

Superskyyy:
> The celery server does not support grpc because it forks server processes and grpc is not good with that so you will need to use the agent http protocol.
> 
> Server script:
> 
> ```python
> #!/usr/bin/env python3
> 
> # docker run --name=redis-devel --publish=6379:6379 --hostname=redis --restart=on-failure --detach redis:latest
> # celery -A t_celery worker -l info
> 
> from celery import Celery
> from time import sleep
> 
> from skywalking import agent
> agent.start()
> 
> app = Celery('test', broker = 'redis://localhost:6379', backend = 'redis://localhost:6379')
> 
> @app.task
> def reverse(text):
>     sleep(1)
>     return text[::-1]
> ```
> 
> Setup:
> 
> ```
> pip install redis celery "apache-skywalking[http]"
> export SW_AGENT_PROTOCOL=http
> export SW_AGENT_COLLECTOR_BACKEND_SERVICES="127.0.0.1:12800"
> docker run --name=redis-devel --publish=6379:6379 --hostname=redis --restart=on-failure --detach redis:latest
> celery -A t_celery worker -l info
> ```
> 
> In the same directory as the server (notice it imports the server module):
> 
> ```python
> #!/usr/bin/env python3
> 
> from skywalking import agent
> agent.start()
> 
> import t_celery
> 
> print(t_celery.reverse.delay('ABC'))
> print(t_celery.reverse('ABC'))
> print(t_celery.reverse.apply_async(('ABC',)))
> print(t_celery.reverse.apply(('ABC',)))
> ```
> 
> In the same directory but another terminal:
> 
> ```
> export SW_AGENT_PROTOCOL=http
> export SW_AGENT_COLLECTOR_BACKEND_SERVICES="127.0.0.1:12800"
> ./t_celery_client.py
> ```
> 
> The final status of the jobs may not be instantaneous to update on the oap server, but all the exit celery requests should connect to to entry requests (except one spurious Local call which I just noticed but don't have time to track down where it came from right now).
> 
> NOTE: Ignore the redis stuff, or just disable the plugin, all you care about is the celery stuff.
> 
> NOTE2: Correct fork handling only works with py 3.7+ since 3.6 does not have the hook, so not sure how http will work on that one (or if).

Thank you! I will look into this.

PR #180, opened 2022-01-30
add fastapi

katelei6:


katelei6:
@kezhenxu94 hello，we want to add fastapi, so everything seems ok, but the response can not catch , could you git us an point to get the response code? the other span seems work ok. Thanks!

skywalking/plugins/sw_fastapi.py
```diff
@@ -0,0 +1,79 @@
...
+    from fastapi.routing import APIRoute
+
+    _get_route_handler = routing.request_response
+
+    def params_tostring(params):
+        return '\n'.join([f"{k}=[{','.join(params.getlist(k))}]" for k, _ in params.items()])
+
+    def _sw_get_route_handler(self, scope: Scope, receive: Receive, send: Send):
+        from starlette.requests import Request
+        print("bububu")
```
kezhenxu94, line 45: Remember to remove all these `print` lines

skywalking/plugins/sw_fastapi.py
```diff
@@ -0,0 +1,79 @@
...
+            span.tag(TagHttpMethod(method))
+            span.tag(TagHttpURL(req.url._url.split('?')[0]))
+            if config.fastapi_collect_http_params and req.values:
+                span.tag(TagHttpParams(params_tostring(req.values)[0:config.http_params_length_threshold]))
+            resp = _get_route_handler(self, scope, receive, send)
+            print(resp)
+            #if isawaitable(resp):
+                #result = await resp
+            # if resp.status_code >= 400:
+            #     span.error_occurred = True
```
kezhenxu94, line 73: Did you investigate the source codes of `routing.request_response`, if it can return both `async` function and normal function, then the commented codes should work

skywalking/plugins/sw_fastapi.py
```diff
@@ -0,0 +1,79 @@
...
+        for item in carrier:
+            if item.key.capitalize() in req.headers:
+                item.val = req.headers[item.key.capitalize()]
+
+        span = NoopSpan(NoopContext()) if config.ignore_http_method_check(method) \
+             else get_context().new_entry_span(op=dict(scope)["path"], carrier=carrier, inherit=Component.General)
+
+        with span:
+            span.layer = Layer.Http
+            span.component = Component.Flask
```
kezhenxu94, line 62: Should have an individual component for FastAPI?

kezhenxu94 (commented):
Please also add the component to main repo like you did for mysqlclient

kezhenxu94:
And try to learn some basic git operations to rebase your new PR on the latest master branch, the commit history is too messy

skywalking/plugins/sw_fastapi.py
katelei6: I see this method， but the args is func, and app is inner function, I did  not try it successful, do not know how to use the inner funncation. I try it again!

PR #181, opened 2022-01-30
Add plugin for FastAPI

katelei6:


katelei6:
@kezhenxu94 we has catch ip,port,path to catch, but the response can not catch. we detect the function is "handle()"， request_response I do not know how to use to get the response? could you git me some idea?

kezhenxu94:
`APIRoute.handle` doesn't return anything, but it has parameters like `receive`, `send` (`Callable` type), is `receive` a callback that will be invoked when there is response? If so, you can wrap `receive` with a custom function and get the response in that callback,

skywalking/plugins/sw_fastapi.py
```diff
@@ -0,0 +1,73 @@
...
+        span = NoopSpan(NoopContext()) if config.ignore_http_method_check(method) \
+            else get_context().new_entry_span(op=dict(scope)["path"], carrier=carrier, inherit=Component.General)
+
+        with span:
+            span.layer = Layer.Http
+            span.component = Component.Flask
+            span.peer = f"{req.client.host}:{req.client.port}"
+            span.tag(TagHttpMethod(method))
+            span.tag(TagHttpURL(req.url._url.split('?')[0]))
+            if config.fastapi_collect_http_params and req.values:
```
wu-sheng, line 62: Where is `fastapi_collect_http_params` declared? I can't find on in the config.

wu-sheng: Also, https://github.com/apache/skywalking-python/blob/master/docs/en/setup/EnvVars.md should be updated accordingly.

katelei6: request_response has not the request args ，this function need function as args,  inner app has the request args, but is the app args can use？

wu-sheng: I just mean the parameter of `fastapi_collect_http_params` for the agent user. 
I am not a qualified reviewer for the Python agent, so, I could be wrong. Mostly, I am just chasing for a good document.

kezhenxu94: > request_response has not the request args ，this function need function as args,  inner app has the request args, but is the app args can use？

I think you are replying the other question. When you want to reply please make sure to reply the right context. Go to the message that you want to reply and click the menu ... in the top-right and choose "quote and reply". Otherwise people may get very confused when reading your comments.

katelei6:
> `APIRoute.handle` doesn't return anything, but it has parameters like `receive`, `send` (`Callable` type), is `receive` a callback that will be invoked when there is response? If so, you can wrap `receive` with a custom function and get the response in that callback,
 katelei6 4 hours ago
request_response has not the request args ，this function need function as args, inner app has the request args, but is the app args can use？I want to try it these day

Superskyyy:
@katelei6 Please update FastAPI to docs/en/setup/Plugins.md by running the tools/doc/plugin_doc_gen.py, this will generate an entry for it based on the plugin versions you specified.

wu-sheng:
@katelei6 Any update?

katelei6:
> Please update FastAPI to docs/en/setup/Plugins.md by running the tools/doc/plugin_doc_gen.py, this will generate an entry for it based on the plugin versions you specified.

katelei6:
I

> @katelei6 Any update?

I will update these days.

katelei6:
> `APIRoute.handle` doesn't return anything, but it has parameters like `receive`, `send` (`Callable` type), is `receive` a callback that will be invoked when there is response? If so, you can wrap `receive` with a custom function and get the response in that callback,

we can not get the response from the `receive`, `send`,

katelei6:
we try some method to get the response for status_code, but was failed. We try it by "receive, send (Callable type)", and request_response method... can you give us some help to get the status_code? we want to use the plugin in our project, but we all java developer, so some source code by python can not understand very well. The tests we can try it, but the response can you give us some help? Or we invoke the method for handle is not the right?
@kezhenxu94

nongyue:
> we try some method to get the response for status_code, but was failed. We try it by "receive, send (Callable type)", and request_response method... can you give us some help to get the status_code? we want to use the plugin in our project, but we all java developer, so some source code by python can not understand very well. The tests we can try it, but the response can you give us some help? Or we invoke the method for handle is not the right? @kezhenxu94

@wu-sheng

wu-sheng:
@nongyue Why ping me?

nongyue:
we have a problem as said as above, we can't figure it out

wu-sheng:
> we have a problem as said as above, we can't figure it out

Then? Why ping me? 
The open-source community is not working in this way. Someone may help, but it takes time. You just posted your question in 2 hours? The people usually are busy with different things or company stuff.

katelei6:
I got it. Has update and get the code! I will add some code and push it. @wu-sheng  @kezhenxu94

skywalking/plugins/sw_fastapi.py
```diff
@@ -0,0 +1,73 @@
...
+                span.tag(TagHttpParams(params_tostring(req.values)[0:config.http_params_length_threshold]))
+            res = await _handle(self, scope, receive, send)
+            # if isawaitable(resp):
+            #     result = await resp
+            # if resp.status_code >= 400:
+            #     span.error_occurred = True
+
+            # span.tag(TagHttpStatusCode(resp.status_code))
```
kezhenxu94, line 70: if `_handle` doesn't return the response, then one of the method arguments might be a callback (or similar) that will be invoked when the response is ready, I don't know why you said "we can't get the response", from what I read, the `send` method looks like a callable method that will be invoked with response as its argument when the response is ready.

If not, you might intercept the wrong method, please learn the FastAPI framework and intercept a proper method, we can be familiar with every framework and we don't know which method you should intercept exactly.

katelei6: The tests failed. The log is "ERROR: for consumer  Container "c1fdeae71565" is unhealthy.", how can I get the detail info?

katelei6:
where can i see the detail log？Our unit tests have error，so we should to edit it, but in our local docker, the unit tests are right!

wu-sheng:
Click here

image

Then you could check the logs from right top

image

katelei6:
How can I curl the url to test? In our local docker, the unit tests is right, but here, I want to try to find how error occur, so I want to curl the url to test if it work right?

wu-sheng:
Could you run this test locally first? You have caused a lot of upstream CI resources and tried repeatedly. This is not a good practice.

tests/plugin/web/sw_fastapi/docker-compose.yml
```diff
@@ -0,0 +1,68 @@
...
+
+  consumer:
+    extends:
+      service: agent
+      file: ../../docker-compose.base.yml
+    ports:
+      - 9090:9090
+    volumes:
+      - .:/app
+    command: ['bash', '-c', 'pip3 install fastapi && pip3 install -r /app/requirements.txt && sw-python run python3 /app/services/consumer.py']
```
Superskyyy, line 55: The same goes to this service.

tests/plugin/web/sw_fastapi/expected.data.yml
```diff
@@ -0,0 +1,73 @@
...
+            tags:
+              - key: http.method
+                value: GET
+              - key: http.url
+                value: http://provider:9091/users
+              - key: http.status.code
+                value: '200'
+            refs:
+              - parentEndpoint: /users
+                networkAddress: 'provider:9091'
```
Superskyyy, line 37: This is not the port your provider service listens, for the sake of consistency, use 9091 as other plugins.

Superskyyy (changes requested):

tests/plugin/web/sw_fastapi/docker-compose.yml
```diff
@@ -0,0 +1,68 @@
...
+
+  provider:
+    extends:
+      service: agent
+      file: ../../docker-compose.base.yml
+    ports:
+      - 9092:9092
+    volumes:
+      - .:/app
+    command: ['bash', '-c', 'pip3 install fastapi && pip3 install -r /app/requirements.txt && sw-python run python3 /app/services/provider.py']
```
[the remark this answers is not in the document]
Superskyyy: image
This fix should work.

tests/plugin/web/sw_fastapi/docker-compose.yml
```diff
@@ -0,0 +1,68 @@
...
+    ports:
+      - 9091:9091
+    volumes:
+      - .:/app
+    command: ['bash', '-c', 'pip3 install fastapi && pip3 install uvicorn && pip3 install -r /app/requirements.txt && sw-python run python3 /app/services/provider.py']
+    depends_on:
+      collector:
+        condition: service_healthy
+    healthcheck:
+      test: ["CMD", "bash", "-c", "cat < /dev/null > /dev/tcp/127.0.0.1/9092"]
```
Superskyyy, line 39: This is why the CI still fails at the beginning.
```suggestion
      test: ["CMD", "bash", "-c", "cat < /dev/null > /dev/tcp/127.0.0.1/9091"]
```

Superskyyy (changes requested):

skywalking/plugins/sw_fastapi.py
```diff
@@ -0,0 +1,79 @@
...
+            else get_context().new_entry_span(op=dict(scope)['path'], carrier=carrier, inherit=Component.General)
+
+        with span:
+            span.layer = Layer.Http
+            span.component = Component.FastApi
+            span.peer = f'{req.client.host}:{req.client.port}'
+            span.tag(TagHttpMethod(method))
+            span.tag(TagHttpURL(req.url._url.split('?')[0]))
+            if config.fastapi_collect_http_params and req.values:
+                span.tag(TagHttpParams(params_tostring(req.values)[0:config.http_params_length_threshold]))
```
Superskyyy: This should be something like `req.query_params` right? `req` is an object of type `MutableMapping(Request)` which has the method `values` inherited. Your code is trying to parse a function instead of the actual query parameter dict, that's why it fails.

katelei6: I change it params of request

katelei6: The assert result not null is wrong?i feel it right but the tests is failed.

Superskyyy: The assert not null is not wrong, it just is a way of showing diffs, the actual wrong thing is the `-` part where no `+` can be matched with, in this case, you are missing the expected entry span in the consumer service. Take another plugin test `expected.data.yml` as a reference you will find what is missing I believe.

katelei6:
I have a question about the docker-compose.yml. First time, the cmd is ‘pip3 install fastapi && pip3 install -r /app/requirements.txt’， the error info is no module of uvicorn, but in the requirements, I add the uvicorn. fastapi==0.70.1
uvicorn==0.16.0
gunicorn==20.1.0, so why need the 'pip3 install uvicorn' too?

kezhenxu94:
> I have a question about the docker-compose.yml. First time, the cmd is ‘pip3 install fastapi && pip3 install -r /app/requirements.txt’， the error info is no module of uvicorn, but in the requirements, I add the uvicorn. fastapi==0.70.1
> uvicorn==0.16.0
> gunicorn==20.1.0, so why need the 'pip3 install uvicorn' too?

@katelei6 the `requirements.txt` in the codebase is to help developers to set up the development environment, not for testing.

For dependencies in tests, you need to add them in `docker-compose.yaml`

katelei6:
I see, thank you for your answer.

katelei6:
I will update the repo ”FastApi:
  id: 7014
  languages: Python“.

tests/plugin/web/sw_fastapi/expected.data.yml
```diff
@@ -0,0 +1,90 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+segmentItems:
+  - serviceName: provider
+    segmentSize: 1
```
wu-sheng, line 20: @kezhenxu94 Is this checked? AFAIK, segment size should be set as `gt 1` or something like this. Because we can't be certain about the size, and the count of following segments is clearly over 1.

Superskyyy:
oO, probably Alpine repo site down, will merge after CI returns to normal.

wu-sheng:
Yes, it seems down. Let's wait for it.

Superskyyy:
CI passed, all good.

PR #182, opened 2022-02-07
Fix sw-rabbitmq TypeError when there are no headers

dcryans:
Add a try/except on TypeError
to catch None headers
when messages in the queue were produced pre-skywalking & without header

Fix [#8499](https://github.com/apache/skywalking/issues/8499)


wu-sheng:
Please fix the lint.

wu-sheng (approved):

dcryans:
Thank you kezhenxu94, sorry didn't use the right linter

PR #185, opened 2022-02-17
Support CI container logs

Superskyyy:
- Supports testcontainer logs from GitHub actions, so no mysterious failures again. 
image

- Also added timeouts to all calls to requests.get/post so that a bad testcase wouldn't hang the entire CI till timeout.
 
- Disabled profiling reporter in test suite since the agent test tool doesn't support the data and keeps throwing irrelevant exceptions.

- Bumped up data protocol dependency to latest master.


wu-sheng (approved):
Please add another changelog, you bumped up the protocol dependency to the latest. (No manual code change required, but the actual running codes are different now)

PR #186, opened 2022-02-23
Update mySQL plugin to support two different parameter keys.

katelei6:
add database param.

closes https://github.com/apache/skywalking/issues/8581

### DIFF
--- a/skywalking/plugins/sw_mysqlclient.py
+++ b/skywalking/plugins/sw_mysqlclient.py
@@ -37,7 +37,9 @@ def install():
     def _sw_connect(*args, **kwargs):
         con = _connect(*args, **kwargs)
         con.host = kwargs['host']
-        con.db = kwargs['db']
+        if "db" in kwargs:
+            con.db = kwargs['db']
+        con.db = kwargs['database']
         return ProxyConnection(con)
 
     class ProxyCursor(wrapt.ObjectProxy):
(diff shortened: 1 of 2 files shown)


skywalking/plugins/sw_mysqlclient.py
```diff
@@ -37,7 +37,9 @@ def install():
     def _sw_connect(*args, **kwargs):
         con = _connect(*args, **kwargs)
         con.host = kwargs['host']
-        con.db = kwargs['db']
+        if 'db' in kwargs:
```
wu-sheng, line 40: If there is no `db` in kwargs, what should we keep in the database name? I think if this is missing, the database would not show up in the database dashboard.

wu-sheng: Why are some in the `database` and the others in `bd`?

katelei6: I  think is the mysqlclient version. I  read the sourse code the lastast version is database. before is db.the connect method handel it like this to avoid version.

katelei6:
The other code I do not update, but the tests is failed.@kezhenxu94

skywalking/plugins/sw_mysqlclient.py
katelei6: I think is the mysqlclient version reason. In lastast, the param is database, but the other is db. so the source in mysqlclient to compatible.

wu-sheng (approved):

PR #187, opened 2022-02-28
add the code > 400 is error

katelei6:
Fix https://github.com/apache/skywalking/issues/8605


wu-sheng (approved):

kezhenxu94 (approved):

katelei6:
@kezhenxu94 the config about the ignore trace is not work well, I set it in env param, but it not work.

kezhenxu94:
> @kezhenxu94 the config about the ignore trace is not work well, I set it in env param, but it not work.

I saw that issue, let's focus on one thing at a time, if you have more information to share, comment in that issue

PR #190, opened 2022-03-18
The local log stack depth is not truncated

wzy960520:


wu-sheng:
@kezhenxu94 https://docs.celeryproject.org/ is really down.

kezhenxu94 (approved):

Superskyyy:
FYI https://github.com/apache/skywalking/issues/8699#issuecomment-1072929993

PR #191, opened 2022-03-19
Revert "The local log stack depth is not truncated (#190)"

Superskyyy:
This reverts commit cd559e7159b92b0d5b9d26cda1da0517a646a709.

Discussion of the reason is hosted > https://github.com/apache/skywalking/issues/8699#issuecomment-1072929993

### DIFF
--- a/skywalking/log/sw_logging.py
+++ b/skywalking/log/sw_logging.py
@@ -69,8 +69,6 @@ def install():
 
         context = get_context()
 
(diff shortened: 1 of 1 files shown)


Superskyyy:
Btw, the Celery website might be totally gone due to unknown ownership of the domain. I will change the link to prevent potential security consequences.

wu-sheng (approved):

PR #195, opened 2022-03-25
Drop support for Flask 1.x due to EOL & Jinja2 issue

Superskyyy:
Flask 1.x is EOL and Jinja2 dependency now breaks UT, there's no need for us to maintain this version anymore.

Please merge this first.

Signed-off-by: Superskyyy <superskyyy@outlook.com>


wu-sheng:
When was its EOL?

Superskyyy:
> When was its EOL?

According to Flask core team's reply, they only support the latest version which is almost 2.1.x. The last 1.x version was released in early 2021 and most users have upgraded to 2.x. 

The main reason is we have to pin dependencies of Flask individually, which is hard to manage with requirements.txt here, anyway that's another story, I should upgrade the project dependency tool to poetry/pdm in the future.

wu-sheng:
Then you need to remove plugins.md accordingly. I am not sure who python agent generates changelog or manually, but we need to highlight we dropped versions.

Superskyyy:
> Then you need to remove plugins.md accordingly. I am not sure who python agent generates changelog or manually, but we need to highlight we dropped versions.

My bad I forgot to regenerate the docs here. I have another PR to update all missing CHANGELOGs coming soon.

wu-sheng (approved):

PR #200, opened 2022-03-30
Support log reporter safe mode

Superskyyy:
- Safe mode involves an additional config that turns on the filtering for HTTP basic auth in log reporter and trace logs (it was originally on by default, now off by default due to REGEX performance in long log records)
- Refactored `config.py` to remove excess expressions mostly like below -
Please correct if I'm wrong, but these and's don't seem necessary 
```
flask_collect_http_params = True if os.getenv('SW_FLASK_COLLECT_HTTP_PARAMS') and \
                                    os.getenv('SW_FLASK_COLLECT_HTTP_PARAMS') == 'True' else False
```
to
```
flask_collect_http_params = os.getenv('SW_FLASK_COLLECT_HTTP_PARAMS') == 'True'
```
- Refactored `config.py` type hints to modern style (Py3.6+ style) 
- Add a e2e entry to verify the log reporter safe mode.
- Attempt to fix the main branch failure situation (I guess it's a correct solution).

Signed-off-by: Superskyyy <superskyyy@outlook.com>


skywalking/config.py
```diff
@@ -17,77 +17,73 @@
...
+# Plugin configurations
+sql_parameters_length: int = int(os.getenv('SW_SQL_PARAMETERS_LENGTH') or '0')
+pymongo_trace_parameters: bool = True if os.getenv('SW_PYMONGO_TRACE_PARAMETERS') == 'True' else False
+pymongo_parameters_max_length: int = int(os.getenv('SW_PYMONGO_PARAMETERS_MAX_LENGTH') or '512')
+elasticsearch_trace_dsl: bool = True if os.getenv('SW_ELASTICSEARCH_TRACE_DSL') == 'True' else False
+
+http_params_length_threshold: int = int(os.getenv('SW_HTTP_PARAMS_LENGTH_THRESHOLD') or '1024')
+http_ignore_method: str = os.getenv('SW_HTTP_IGNORE_METHOD', '').upper()
+flask_collect_http_params: bool = True if os.getenv('SW_FLASK_COLLECT_HTTP_PARAMS') == 'True' else False
+sanic_collect_http_params: bool = True if os.getenv('SW_SANIC_COLLECT_HTTP_PARAMS') == 'True' else False
```
kezhenxu94, line 62: This kind of `bool` expression can be simplified further?

```suggestion
sanic_collect_http_params: bool = os.getenv('SW_SANIC_COLLECT_HTTP_PARAMS') == 'True'
```

kezhenxu94 (approved):

skywalking/config.py
Superskyyy: You r right, 1sec to fix them.

kezhenxu94 (approved):

PR #205, opened 2022-04-26
Remove namespace to instance properties and add pid

kezhenxu94:
Closes https://github.com/apache/skywalking/issues/8665


wu-sheng (approved):

Superskyyy (approved):
lgtm

PR #208, opened 2022-05-24
Unify the tag name with other agents

kezhenxu94:


wu-sheng (approved):

wu-sheng:
I think test cases break?

Superskyyy (approved):
LGTM

PR #210, opened 2022-05-30
Improved ignore path regex

tom-pytel:
This is the same as skywalking-nodejs PR #81:

Minor enhancement/fix? Previously the config.trace_ignore_path pattern `/a/**/*` would match `/a/b/`, `/a/b/c` and `/a/b/c/`, etc... but not `/a/` or `/a/b`. Now it does.

### DIFF
--- a/skywalking/config.py
+++ b/skywalking/config.py
@@ -104,12 +104,14 @@ def finalize():
     method = r'^' + '|'.join(s.strip() for s in http_ignore_method.split(',')) + '$'
     path = '^(?:' + \
            '|'.join(  # replaces ","
-               '(?:(?:[^/]+/)*[^/]+)?'.join(  # replaces "**"
-                   '[^/]*'.join(  # replaces "*"
-                       '[^/]'.join(  # replaces "?"
-                           reesc.sub(r'\\\1', s) for s in p2.split('?')
-                       ) for p2 in p1.split('*')
-                   ) for p1 in p0.strip().split('**')
+               '/(?:[^/]*/)*'.join(  # replaces "/**/"
+                   '(?:(?:[^/]+/)*[^/]+)?'.join(  # replaces "**"
+                       '[^/]*'.join(  # replaces "*"
+                           '[^/]'.join(  # replaces "?"
+                               reesc.sub(r'\\\1', s) for s in p3.split('?')
+                           ) for p3 in p2.split('*')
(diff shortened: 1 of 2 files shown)


tom-pytel:
Update test case, `/**/` should match any number of path segments OR NONE, test case was expecting to match at least one. Correct pattern to match AT LEAST ONE is `/*/**/`.

Superskyyy:
I think that a `/a/**/*` pattern should not match `/a/` because there must be another `/` after the `**`. 

Just want to point out the potential mismatch with a common perception of glob syntax.

tom-pytel:
> I think that a `/a/**/*` pattern should not match `/a/` because there must be another `/` after the `**`.
> 
> Just want to point out the potential mismatch with a common perception of glob syntax. What do you think?

Could be mismatch with standard glob but the idea is to allow `/**/*` to be a wildcard that matches anything, including `/`, `/a`, `/a/`, `/a/b`, etc... Otherwise it won't and you will need two patterns. Your call, if you want I can revert to standard behavior in the node agent as well.

Superskyyy:
I think you are right, this is indeed more robust. Probably adding a small FAQ documentation will be nice, so it eliminates user confusion.

Superskyyy:
Other than this concern all LGTM.

tom-pytel:
> I think you are right, this is indeed more robust. Probably adding a small FAQ documentation will be nice, so it eliminates user confusion.

I was looking for a place to document this and found the following in `/docs/en/setup/EnvVars.md` the following:

"SW_TRACE_IGNORE_PATH: You can setup multiple URL path patterns, The endpoints match these patterns wouldn't be traced. the current matching rules follow `Ant Path match style` , like /path/*, /path/**, /path/?"

Then, looking up `Ant Path match style`:

`**` matches zero or more 'directories' in a path

And:

`com/**/test.jsp` - matches all `test.jsp` files underneath the `com` path

And:

`org/springframework/**/*.jsp` - matches all `.jsp` files underneath the `org/springframework path`

Which implies the previous behavior and test below did not actually follow the specification. Since now it does, there is nothing extra to document I think?
```
pattern = 'eureka/**/test/**'
path = 'eureka/test/list'
self.assertFalse(fast_path_match(pattern, path))
```

Superskyyy:
I see, I didn't take a careful look ealier. Then everything LGTM :)

Superskyyy (approved):

PR #213, opened 2022-06-13
Add plugin doc check in CI

JarvisG495:
- Add plugin doc check in CI
- Fix a dead link
- Replace `tools/doc/plugin_doc_gen.py` with `make doc-gen` in contribution docs

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,7 @@ doc-gen:
 	python3 -m tools.doc.plugin_doc_gen
 
 check-doc-gen:
-	$(MAKE) install
+	$(MAKE) setup-test
 	$(MAKE) doc-gen
 	@if [ ! -z "`git status -s`" ]; then \
 		echo "Plugin doc is not consisitent with CI:"; \
--- a/Makefile
+++ b/Makefile
@@ -25,6 +25,9 @@ endif
 
 .PHONY: license
 
+$(VENV):
+	python3 -m venv $(VENV_DIR)
+
 setup:
 	python3 -m pip install --upgrade pip
 	python3 -m pip install grpcio --ignore-installed
@@ -43,6 +46,7 @@ lint: clean
 
 # used in development
 dev-setup:
+	$(VENV)
 	$(VENV)/python -m pip install -r requirements-style.txt
 
 dev-check: dev-setup
@@ -55,9 +59,10 @@ dev-fix: dev-setup
 	$(VENV)/flynt -tc -v .
 
 doc-gen:
-	python3 -m tools.doc.plugin_doc_gen
+	$(VENV)/python tools/doc/plugin_doc_gen.py
 
 check-doc-gen:
+	$(VENV)
 	$(MAKE) setup-test
 	$(MAKE) doc-gen
 	@if [ ! -z "`git status -s`" ]; then \
--- a/Makefile
+++ b/Makefile
@@ -58,7 +58,6 @@ doc-gen:
 	$(VENV)/python tools/doc/plugin_doc_gen.py
 
 check-doc-gen:
-	$(MAKE) clean
 	$(VENV)/python tools/doc/plugin_doc_gen.py
 	@if [ ! -z "`git status -s`" ]; then \
 		echo "Plugin doc is not consisitent with CI:"; \
--- a/Makefile
+++ b/Makefile
@@ -27,22 +27,22 @@ endif
 
 $(VENV):
 	python3 -m venv $(VENV_DIR)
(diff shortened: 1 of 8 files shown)


Makefile
```diff
@@ -55,7 +55,16 @@ dev-fix: dev-setup
 	$(VENV)/flynt -tc -v .
 
 doc-gen:
-	$(VENV)/python tools/doc/plugin_doc_gen.py
+	python3 -m tools.doc.plugin_doc_gen
```
kezhenxu94, line 58: can we keep using `$(VENV)/python`?

JarvisG495: Yes. I made this change only because CI environment does not set up venv and check-doc-gen needs doc-gen.
Maybe we can keep using `$(VENV)/python` in doc-gen and use `python3 -m tools.doc.plugin_doc_gen` instead of `make doc-gen` in check-doc-gen.
What do you think?

kezhenxu94: @JarvisG495 let's add a goal in Makefile

```makefile
$(VENV):
	python3 -m venv $(VENV_DIR)
```

And add dependent `$(VENV)` to all other goals that use `$(VENV)`, so that it `venv` can  be initialized automatically

Makefile
```diff
@@ -43,6 +46,7 @@ lint: clean
 
 # used in development
 dev-setup:
+	$(VENV)
```
kezhenxu94, line 49: This 👇 is the correct form of "dependent"

```suggestion
dev-setup: $(VENV)
```

JarvisG495: Oh, I see. Corrected.

kezhenxu94 (approved):
Thank you @JarvisG495

Superskyyy:
It seems like CI wouldn't pass, let's take a look

Superskyyy:
````
ERROR: for 2488337382_provider_1  Cannot start service provider: failed to create shim: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: "sw-python": executable file not found in $PATH: unknown
```

The testing env is using the wrong Python environment, introduced by venv, let's fix it.

Superskyyy:
Very nice, LGTM, thank you.

Superskyyy:
Let's try move these lines to below license-and-lint, also add this job name as a prerequisite of further ones like what we did in license and lint, so we can fail fast if these basic checks fail.
https://github.com/apache/skywalking-python/blob/f430e46c7e8f1edba281f46d174d4dfd8ae35f0f/.github/workflows/CI.yaml#L101-L115

JarvisG495:
> Let's try move these lines to below license-and-lint, also add this job name as a prerequisite of further ones like what we did in license and lint, so we can fail fast if these basic checks fail.
> 
> https://github.com/apache/skywalking-python/blob/f430e46c7e8f1edba281f46d174d4dfd8ae35f0f/.github/workflows/CI.yaml#L101-L115

np, and I think these should be moved to the part right below `changes`, since it can be skipped if no changes are found, right?

Superskyyy:
> > Let's try move these lines to below license-and-lint, also add this job name as a prerequisite of further ones like what we did in license and lint, so we can fail fast if these basic checks fail.
> 
> > 
> 
> > https://github.com/apache/skywalking-python/blob/f430e46c7e8f1edba281f46d174d4dfd8ae35f0f/.github/workflows/CI.yaml#L101-L115
> 
> 
> 
> np, and I think these should be moved to the part right below `changes`, since it can be skipped if no changes are found, right?

Imagine when you accidentally/ unknowingly change the plugin documentation, it will not trigger the changes job because it's in the docs folder, thus skipping the check, but this shouldn't happen right?

JarvisG495:
> > > Let's try move these lines to below license-and-lint, also add this job name as a prerequisite of further ones like what we did in license and lint, so we can fail fast if these basic checks fail.
> > 
> > 
> > > 
> > 
> > 
> > > https://github.com/apache/skywalking-python/blob/f430e46c7e8f1edba281f46d174d4dfd8ae35f0f/.github/workflows/CI.yaml#L101-L115
> > 
> > 
> > np, and I think these should be moved to the part right below `changes`, since it can be skipped if no changes are found, right?
> 
> Imagine when you accidentally/ unknowingly change the plugin documentation, it will not trigger the changes job because it's in the docs folder, thus skipping the check, but this shouldn't happen right?

Make sense. I missed that case. I have updated `plugin-doc-check` location and also modified the exit codes in `CheckStatus` accordingly to follow the execution order.

Superskyyy:
Looks great.

Superskyyy (approved):

PR #214, opened 2022-06-27
Add plugin for bottle

jiang1997:
- [x] Add a test case for the new plugin
- [x] Add a CHANGELOG entry for the new plugin
- [x] Rebuild the `Plugins.md` documentation by running `make doc-gen`


Superskyyy:
Thanks for the contribution, please fix the lint and update your base branch.

jiang1997:
my fault 😂

Superskyyy:
> my fault 😂
Not a problem！
Please do a reformat if you are using pycharm for files **you added**, it will fix most things.

Superskyyy:
Please also add a entry in the Changelog thanks

skywalking/plugins/sw_bottle.py
```diff
@@ -0,0 +1,68 @@
...
+from skywalking import Layer, Component, config
+from skywalking.trace.carrier import Carrier
+from skywalking.trace.context import get_context, NoopContext
+from skywalking.trace.span import NoopSpan
+from skywalking.trace.tags import TagHttpMethod, TagHttpURL
+
+link_vector = ['http://bottlepy.org/docs/dev/']
+support_matrix = {
+    'bottle': {
+        '>=0.12': ['*']
```
Superskyyy, line 27: This isn't a supported syntax, and you confused the left and right side. 

left side should be version of Python right side is the package, see and example below.
```
support_matrix = {
    'hug': {
        '>=3.10': ['2.5', '2.6'],  # api deprecated for 3.10
        '>=3.6': ['2.4.1', '2.5', '2.6'],  # support begins 2.4.1
    }
}
```

docs/en/setup/Plugins.md
```diff
@@ -14,6 +14,7 @@ or a limitation of SkyWalking auto-instrumentation (welcome to contribute!)
 Library | Python Version - Lib Version | Plugin Name
 | :--- | :--- | :--- |
 | [aiohttp](https://docs.aiohttp.org) | Python >=3.10 - NOT SUPPORTED YET; Python >=3.6 - ['3.7.4'];  | `sw_aiohttp` |
+| [bottle](http://bottlepy.org/docs/dev/) | Python >=0.12 - ['*'];  | `sw_bottle` |
```
Superskyyy, line 17: see here you can tell the generated doc isn't correct.

jiang1997: got it.

jiang1997: Is there any other error reported in Actions? I want to eliminate as many mistakes as possible on each commit. Thanks!

wu-sheng: It is better you could run the tests locally. Then you could catch all errors.

jiang1997: > It is better you could run the tests locally. Then you could catch all errors.

Yea, I make it now.
Here is the problem I encountered.
```
docker pull apache/skywalking-python-agent:latest-plugin
```
The above command always runs into trouble with the error message below
```
Error response from daemon: pull access denied for apache/skywalking-python-agent, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
```
Seems the latest tag is unavailable, and docker-compose.base.yml under the tests/plugin folder needs to be altered for the test in my situation.

jiang1997: > > Is there any other error reported in Actions? I want to eliminate as many mistakes as possible on each commit. Thanks!
> 
> not sure, but seems like you are almost there, remember to regenerate docs and add a changelog

bear in mind

jiang1997 (commented):
The unit test refers to Django's.

docs/en/setup/Plugins.md
kezhenxu94: > > It is better you could run the tests locally. Then you could catch all errors.
> 
> Yea, I make it now. Here is the problem I encountered.
> 
> ```
> docker pull apache/skywalking-python-agent:latest-plugin
> ```
> 
> The above command always runs into trouble with the error message below
> 
> ```
> Error response from daemon: pull access denied for apache/skywalking-python-agent, repository does not exist or may require 'docker login': denied: requested access to the resource is denied
> ```
> 
> Seems the latest tag is unavailable, and docker-compose.base.yml under the tests/plugin folder needs to be altered for the test in my situation.

Sorry for my belated response. This Docker image is built on the fly right before running the tests. Please refer to https://github.com/apache/skywalking-python/blob/88b30ae094079167409e7c2bf4ff68431bc9e96a/.github/workflows/CI.yaml#L135

kezhenxu94: > Is there any other error reported in Actions? I want to eliminate as many mistakes as possible on each commit. Thanks!

Hi. Can you see the actions log [here](https://github.com/apache/skywalking-python/runs/7165080924?check_suite_focus=true#step:6:177)? There is a diff output that indicates what’s the expected data and what’s the actual data (ignore the matching pattern like `gt 0`, `not null`, etc.)

jiang1997 (commented):
Here is the way I create "apache/skywalking-python-agent:latest-plugin" image for the test.
First, create the file as below,
skywalking-python/Dockerfile.test
```
FROM python:3.8

WORKDIR /usr/src/app

COPY . .

RUN pip install -r requirements.txt
RUN pip install -e .[test]
```
then
```
docker build -t apache/skywalking-python-agent:latest-plugin .  -f Dockerfile.test 
```

jiang1997:
> 



> Here is the way I create "apache/skywalking-python-agent:latest-plugin" image for the test. First, create the file as below, skywalking-python/Dockerfile.test
> 
> ```
> FROM python:3.8
> 
> WORKDIR /usr/src/app
> 
> COPY . .
> 
> RUN pip install -r requirements.txt
> RUN pip install -e .[test]
> ```
> 
> then
> 
> ```
> docker build -t apache/skywalking-python-agent:latest-plugin .  -f Dockerfile.test 
> ```

The reason  I mention this is that the instruction to run a test locally is kind of vague by now.

kezhenxu94:
> The reason I mention this is that the instruction to run a test locally is kind of vague by now.

I'm sorry to hear that... and I've created [an issue](https://github.com/apache/skywalking/issues/9301) to track this issue, if you're interested in helping to complete the doc, feel free to open a PR and add the guide about running tests locally.

Thanks for add this plugin anyway!

kezhenxu94 (approved):
LGTM, @Superskyyy wanna recheck?

jiang1997:
> > The reason I mention this is that the instruction to run a test locally is kind of vague by now.
> 
> I'm sorry to hear that... and I've created [an issue](https://github.com/apache/skywalking/issues/9301) to track this issue, if you're interested in helping to complete the doc, feel free to open a PR and add the guide about running tests locally.
> 
> Thanks for add this plugin anyway!

I'm glad to go through with these.

wu-sheng:
@jiang1997 I think you missed two steps besides this PR.

1. Add component ID into the main repo.
2. Add logo(if it has one) into booster UI. Ping @kezhenxu94 It seems this PR template is still leading to rocketbot UI.

jiang1997:
> @jiang1997 I think you missed two steps besides this PR.
> 
> 1. Add component ID into the main repo.
> 2. Add logo(if it has one) into booster UI. Ping @kezhenxu94 It seems this PR template is still leading to rocketbot UI.



> @jiang1997 I think you missed two steps besides this PR.
> 
> 1. Add component ID into the main repo.
> 2. Add logo(if it has one) into booster UI. Ping @kezhenxu94 It seems this PR template is still leading to rocketbot UI.

So I am supposed to do those before this merge, right?

wu-sheng:
Yes, at least the component ID should get merged first.

PR #216, opened 2022-07-02
Fix typo in docker/Makefile

jiang1997:


docker/Makefile
```diff
@@ -20,11 +20,13 @@ P := grpc http kafka
 
 TARGETS := py3.6 py3.7 py3.8 py3.9 py3.10
 
+AGENT_VERSION := 0.7
```
wu-sheng, line 23: Don't lock the agent version in CI, this could be left after release.

kezhenxu94, line 23: Hi. This is set  before running the make targets. Did you read this about how to build docker images? https://github.com/apache/skywalking-python/blob/master/docs/en/contribution/How-to-release-docker.md#how-to-build

docker/Makefile
```diff
@@ -20,11 +20,13 @@ P := grpc http kafka
...
+
 py3.6: BASE_PYTHON_IMAGE = python:3.6
 py3.7: BASE_PYTHON_IMAGE = python:3.7
 py3.8: BASE_PYTHON_IMAGE = python:3.8
 py3.9: BASE_PYTHON_IMAGE = python:3.9
-py3.9: BASE_PYTHON_IMAGE = python:3.10
+py3.10: BASE_PYTHON_IMAGE = python:3.10
```
kezhenxu94, line 29: This is a good catch!

wu-sheng (approved):

PR #217, opened 2022-07-04
Update UI repository link in PR template

kezhenxu94:

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE
+++ b/.github/PULL_REQUEST_TEMPLATE
@@ -2,7 +2,7 @@
 <!--
 - [ ] Add a test case for the new plugin
 - [ ] Add a CHANGELOG entry for the new plugin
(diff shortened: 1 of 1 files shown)


.github/PULL_REQUEST_TEMPLATE
```diff
@@ -3,7 +3,7 @@
 - [ ] Add a test case for the new plugin
 - [ ] Add a CHANGELOG entry for the new plugin
 - [ ] Add a component id in [the main repo](https://github.com/apache/skywalking/blob/master/oap-server/server-bootstrap/src/main/resources/component-libraries.yml#L415)
```
jiang1997, line 5: This link seems 404 now. 
Maybe it should be changed to ``https://github.com/apache/skywalking/blob/master/oap-server/server-starter/src/main/resources/component-libraries.yml``

wu-sheng (approved):

PR #222, opened 2022-07-11
doc: add How to test locally

jiang1997:

### DIFF
--- a/docs/en/setup/faq/How-to-test-locally.md
+++ b/docs/en/setup/faq/How-to-test-locally.md
@@ -1,3 +1,5 @@
+# How to test locally?
+
 This guide assumes you just cloned the repo and are ready to make some changes.
 
 After cloning the repo, make sure you also have cloned the submodule for protocol. Otherwise, run the command below. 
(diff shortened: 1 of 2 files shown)


docs/en/setup/faq/How-to-test-locally.md
```diff
@@ -0,0 +1,36 @@
...
+
+RUN pip install -r requirements.txt
+RUN pip install -e .[test]
+then
+```
+
+Then run the command below to create the docker image we need later. Because the test process runs in the docker container, we need to apply the changes you did to the docker image. You may need to run this command whenever you change any files before the test.
+```
+docker build -t apache/skywalking-python-agent:latest-plugin .  -f Dockerfile.test 
+```
```
kezhenxu94, line 34: All these can be simplified as 

```
docker build --build-arg BASE_PYTHON_IMAGE=3.6 -t apache/skywalking-python-agent:latest-plugin --no-cache . -f tests/plugin/Dockerfile.plugin
```

kezhenxu94 (commented):
Thank you @jiang1997 for adding this

docs/en/setup/faq/How-to-test-locally.md
jiang1997: Yea, when users have not run any make command, and there is no venv folder, this one-line command works as expected.
I observed when there is venv folder in ``ROOT``, this one-line command runs into error as below. 
```
Step 6/7 : RUN cd /agent && make setup-test
 ---> Running in 8c58f3791fcb
venv/bin/python -m pip install grpcio --ignore-installed
/agent/venv/bin/python: No module named pip
make: *** [Makefile:33: setup] Error 1
The command '/bin/sh -c cd /agent && make setup-test' returned a non-zero code: 2
```
I guess because the existed venv is not configured for the environment inside docker image, but make script runs inside docker image do not aware of it.

After changing ``tests/plugin/Dockerfile.plugin`` as below, everything works fine.
```
ARG BASE_PYTHON_IMAGE

FROM python:${BASE_PYTHON_IMAGE}

ARG ROOT=.

WORKDIR /agent

ADD $ROOT /agent

RUN cd /agent && rm -rf venv && make setup-test

ENV PATH="/agent/venv/bin:$PATH"
```

kezhenxu94: Hi @jiang1997 thanks for pointing this out. I think you're right. We have another more canonical way to avoid this. Can you please add a file `.dockerignore` in root directory of this project and add the `venv` as its content? This should resolve the problem. 

File `.dockerignore`: 

```
venv
```

kezhenxu94: Don't forget to add license header for the new file

PR #230, opened 2022-08-19
Add aioredis, aiormq, amqp, asyncpg, aio-pika, kombu RMQ

tom-pytel:
Added the following explicit plugins: aioredis, aiormq, amqp, asyncpg.
These implicitly enable aio-pika and kombu rabbitmq to work.
Will add component IDs to main sw repo.

Unfortunately I am not intelligent enough to write the tests for these. If anyone out there with a big brain and a love for testing gets the urge, feel free. I've left some test examples in two of the plugins and the rest should be easy enough to get from their respective sites. Also I only tested with the versions given, someone may want to determine version range of compatibilities.


wu-sheng:
> Unfortunately I am not intelligent enough to write the tests for these.

Are these commercials or cloud services? So it os hard to test?

tom-pytel:
> Are these commercials or cloud services? So it os hard to test?

I was just kidding, nothing hard about it except I never got the hang of the testing system you guys use and don't have so much free time to come up with them. I am essentially backporting these from our fork of SW so maybe some juniors on your side want to contribute something?

wu-sheng:
I see. I will leave this to @kezhenxu94 AFAIK, we don't have ppl specific for python agent, he does all these in his free time only.

Superskyyy:
We have a new contributor from OSPP, maybe he can take on some task to complete the tests (or come up with a better testing strategies, the current is pretty hard to maintain up-to-date)

Superskyyy:
And code LGTM. Thank you!

Superskyyy (approved):

tom-pytel:
For the tests I did it always took me 2-3x as long as the actual thing to be tested.

PR #231, opened 2022-08-22
feat: Add support for MeterReportService

jiang1997:
Add support for MeterReportService

# Behavior
This service runs along with the agent like other services and reports roughly every 20 seconds.

### Counter
```python
c = Counter('c2', CounterMode.INCREMENT)
c.build()

# increase Counter c by the time the with-wrapped codes consumed
with c.create_timer():
    # some codes may consume a certain time
```

```python
c = Counter('c3', CounterMode.INCREMENT)
c.build()

# increase Counter c by num once counter_decorator_test gets called
@Counter.increase(name='c3', num=2)
def counter_decorator_test():
    # some codes
```

```python
c = Counter('c4', CounterMode.INCREMENT)
c.build()

# increase Counter c by the time counter_decorator_test consumed
@Counter.timer(name='c4')
def counter_decorator_test(s):
    # some codes may consume a certain time
```
### Histogram

```python
h = Histogram('h3', [i / 10 for i in range(10)])
h.build()

# Histogram h will record the time the with-wrapped codes consumed
with h.create_timer():
    # some codes may consume a certain time
```

```python
h = Histogram('h2', [i / 10 for i in range(10)])
h.build()

# Histogram h will record the time histogram_decorator_test consumed
@Histogram.timer(name='h2')
def histogram_decorator_test(s):
    time.sleep(s)
```

# Simple test
insert following codes below this line ``agent/__init__.py#L151``
```python
from skywalking.meter.gauge import Gauge
import gc

def generator_for_gc_g0():
    while(True):
        yield gc.get_stats()[0]['collected']

def generator_for_gc_g1():
    while(True):
        yield gc.get_stats()[1]['collected']

def generator_for_gc_g2():
    while(True):
        yield gc.get_stats()[2]['collected']


gc_g0_meter = Gauge("gc_g0", generator_for_gc_g0())
gc_g0_meter.build()

gc_g1_meter = Gauge("gc_g1", generator_for_gc_g1())
gc_g1_meter.build()

gc_g2_meter = Gauge("gc_g2", generator_for_gc_g2())
gc_g2_meter.build()
```

### DIFF
--- a/docs/en/setup/advanced/MeterReporter.md
+++ b/docs/en/setup/advanced/MeterReporter.md
@@ -1,0 +1,110 @@
+# Python Agent Meter Reporter
+
+To enable or disable this feature, you will need to set some environment variables.
+
+
+## Enabling the feature (default)
+```Python 
+os.environ['SW_AGENT_METER_REPORTER_ACTIVE'] = 'True'
+``` 
+or
+```bash
+export SW_AGENT_METER_REPORTER_ACTIVE=True
+```
+
+## Disable the feature
+```Python 
+os.environ['SW_AGENT_METER_REPORTER_ACTIVE'] = 'False'
+``` 
+or
+```bash
+export SW_AGENT_METER_REPORTER_ACTIVE=False
+```
+## Counter
+* `Counter` API represents a single monotonically increasing counter, automatic collect data and report to backend.
+```python
+tg_ls = [MeterTag("key", "value")]
+c = Counter('c2', CounterMode.INCREMENT, tg_ls)
+# or with more compact way
+# Counter('c2', CounterMode.INCREMENT).tag('key1', 'value1').tag('key2', 'value2')
+c.build()
+c.increment(2)
+```
+### Syntactic sugars
+```python
+c = Counter('c2', CounterMode.INCREMENT)
+c.build()
+
+# increase Counter c by the time the with-wrapped codes consumed
+with c.create_timer():
+    # some codes may consume a certain time
+```
+
+```python
+c = Counter('c3', CounterMode.INCREMENT)
+c.build()
+
+# increase Counter c by num once counter_decorator_test gets called
+@Counter.increase(name='c3', num=2)
+def counter_decorator_test():
+    # some codes
+```
+
+```python
+c = Counter('c4', CounterMode.INCREMENT)
+c.build()
+
+# increase Counter c by the time counter_decorator_test consumed
+@Counter.timer(name='c4')
+def counter_decorator_test(s):
+    # some codes may consume a certain time
+```
+
+1. `Counter(name, tags)` Create a new counter with the meter name and optional tags.
+1. `Counter.tag(key: str, value)` Mark a tag key/value pair.
+1. `Counter.mode(mode: CounterMode)` Change the counter mode, RATE mode means reporting rate to the backend.
+1. `Counter.increment(count)` Increment count to the `Counter`, It could be a positive value.
+
+## Gauge
+* `Gauge` API represents a single numerical value.
+```python
+tg_ls = [MeterTag("key", "value")]
+# producer: iterable object
+g = Gauge('g1', producer, tg_ls)
+g.build()
+```
+1. `Gauge(name, tags)` Create a new gauge with the meter name and iterable object, this iterable object need to produce numeric value.
+1. `Gauge.tag(key: str, value)` Mark a tag key/value pair.
+1. `Gauge.build()` Build a new `Gauge` which is collected and reported to the backend.
+
+## Histogram
+* `Histogram` API represents a summary sample observations with customize buckets.
+```python
+tg_ls = [MeterTag("key", "value")]
+h = Histogram('h2', [i / 10 for i in range(10)], tg_ls)
+h.build()
+```
+### Syntactic sugars
+```python
+h = Histogram('h3', [i / 10 for i in range(10)])
+h.build()
+
+# Histogram h will record the time the with-wrapped codes consumed
+with h.create_timer():
+    # some codes may consume a certain time
+```
+
+```python
+h = Histogram('h2', [i / 10 for i in range(10)])
+h.build()
+
+# Histogram h will record the time histogram_decorator_test consumed
+@Histogram.timer(name='h2')
+def histogram_decorator_test(s):
+    time.sleep(s)
+```
+1. `Histogram(name, tags)` Create a new histogram with the meter name and optional tags.
+1. `Histogram.tag(key: str, value)` Mark a tag key/value pair.
+1. `Histogram.minValue(value)` Set up the minimal value of this histogram, default is `0`.
+1. `Histogram.build()` Build a new `Histogram` which is collected and reported to the backend.
+1. `Histogram.addValue(value)` Add value into the histogram, automatically analyze what bucket count needs to be increment. rule: count into [step1, step2).
--- a/skywalking/agent/__init__.py
+++ b/skywalking/agent/__init__.py
@@ -27,6 +27,7 @@ from skywalking import profile
 from skywalking.agent.protocol import Protocol
 from skywalking.command import command_service
 from skywalking.loggings import logger
+from skywalking.meter.meter_service import MeterService
 from skywalking.profile.profile_task import ProfileTask
 from skywalking.profile.snapshot import TracingThreadSnapshot
 from skywalking.protocol.logging.Logging_pb2 import LogData
@@ -37,8 +38,7 @@ if TYPE_CHECKING:
 __started = False
 __protocol = Protocol()  # type: Protocol
 __heartbeat_thread = __report_thread = __log_report_thread = __query_profile_thread = __command_dispatch_thread \
-    = __send_profile_thread = __queue = __log_queue = __snapshot_queue = __finished = None
-
+    = __send_profile_thread = meter_service_thread = __queue = __log_queue = __snapshot_queue = __finished = None
 
 def __heartbeat():
     wait = base = 30
@@ -117,7 +117,7 @@ def __command_dispatch():
 
 def __init_threading():
     global __heartbeat_thread, __report_thread, __log_report_thread, __query_profile_thread, \
-        __command_dispatch_thread, __send_profile_thread, __queue, __log_queue, __snapshot_queue, __finished
+        __command_dispatch_thread, __send_profile_thread, meter_service_thread, __queue, __log_queue, __snapshot_queue, __finished
 
     __queue = Queue(maxsize=config.max_buffer_size)
     __finished = Event()
@@ -128,6 +128,10 @@ def __init_threading():
     __heartbeat_thread.start()
     __report_thread.start()
     __command_dispatch_thread.start()
+
+    if config.meter_reporter_activate:
+        meter_service_thread = MeterService(__protocol.meter_reporter, __finished, logger)
+        meter_service_thread.start()
 
     if config.log_reporter_active:
         __log_queue = Queue(maxsize=config.log_reporter_max_buffer_size)
--- a/skywalking/agent/__init__.py
+++ b/skywalking/agent/__init__.py
@@ -24,13 +24,14 @@ from typing import TYPE_CHECKING
 from skywalking import config, plugins
 from skywalking import loggings
 from skywalking import profile
+from skywalking import meter
 from skywalking.agent.protocol import Protocol
 from skywalking.command import command_service
 from skywalking.loggings import logger
-from skywalking.meter.meter_service import MeterService
 from skywalking.profile.profile_task import ProfileTask
 from skywalking.profile.snapshot import TracingThreadSnapshot
 from skywalking.protocol.logging.Logging_pb2 import LogData
+from skywalking.protocol.language_agent.Meter_pb2 import MeterData
 
 if TYPE_CHECKING:
     from skywalking.trace.context import Segment
@@ -38,7 +39,7 @@ if TYPE_CHECKING:
 __started = False
 __protocol = Protocol()  # type: Protocol
 __heartbeat_thread = __report_thread = __log_report_thread = __query_profile_thread = __command_dispatch_thread \
-    = __send_profile_thread = meter_service_thread = __queue = __log_queue = __snapshot_queue = __finished = None
+    = __send_profile_thread = __queue = __log_queue = __snapshot_queue = __meter_queue = __finished = None
 
 
 def __heartbeat():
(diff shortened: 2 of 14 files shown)


Superskyyy:
Please try to run the linter with `make lint` or `make dev-check`(if you don't want venv egg. to be purged) command. It will be quicker to fix things.

skywalking/client/__init__.py
```diff
@@ -28,6 +28,9 @@ class TraceSegmentReportService(object):
     def report(self, generator):
         raise NotImplementedError()
 
+class MeterReportService(object):
+    def report(self, generator):
```
Superskyyy, line 32: also report_batch?

skywalking/meter/histogram.py
```diff
@@ -0,0 +1,135 @@
...
+
+
+class Histogram(BaseMeter):
+    def __init__(self, name: str, steps, minValue=0, tags=[]):
+        super().__init__(name, tags)
+        self.minValue = minValue
+
+        # https://stackoverflow.com/a/53522/9845190
+        if not steps:
+            # raise error
```
Superskyyy, line 30: Is something missing here? As the comment says - raise an error.

skywalking/meter/histogram.py
```diff
@@ -0,0 +1,135 @@
...
+            pass
+
+        # https://stackoverflow.com/a/2931683/9845190
+        steps = sorted(set(steps))
+
+        if steps[0] < self.minValue:
+            # raise error
```
Superskyyy, line 37: Same as above.

skywalking/meter/meter.py
```diff
@@ -0,0 +1,115 @@
...
+        return self.tags
+    
+    def getType(self):
+        return self.type
+    
+    def __hash__(self):
+        return hash((self.name, self.type, tuple(self.tags)))
+
+
+class BaseMeter():
```
Superskyyy, line 74: Missing a ABC here, otherwise the @abstractmethod won't work as expected

jiang1997:
Sorry for messing up the commit history.

skywalking/agent/__init__.py
```diff
@@ -37,7 +38,7 @@
 __started = False
 __protocol = Protocol()  # type: Protocol
 __heartbeat_thread = __report_thread = __log_report_thread = __query_profile_thread = __command_dispatch_thread \
-    = __send_profile_thread = __queue = __log_queue = __snapshot_queue = __finished = None
+    = __send_profile_thread = meter_service_thread = __queue = __log_queue = __snapshot_queue = __finished = None
```
jiang1997, line 41: I tried name ``meter_service_thread`` as ``__meter_service_thread`` comply the naming convention here.But only to find that double underscore has its own specific behaviour when importing double underscore variable from another file. More details here. https://stackoverflow.com/questions/1301346/what-is-the-meaning-of-single-and-double-underscore-before-an-object-name

Superskyyy:
> Sorry for messing up the commit history.

It's okay, everything will be squashed in the end.

jiang1997:
> > Sorry for messing up the commit history.
> 
> It's okay, everything will be squashed in the end.

Yes, it makes sense for both the end of this pr and the universe.

skywalking/agent/__init__.py
Superskyyy: I believe `__` prefixed variables are not supposed to be referred to by external classes, but when it is defined outside of classes, it doesn't do much other than the meaning of "you probably shouldn't use (import) this". 

Anyway, I see the current implementation bit out of sync with the previous reporter styles. 

That is, you provided a new `MeterService(Thread)`, and the result of defining threads outside of `agent/__init__.py` is that the code base becomes weird in terms of calling topology, and there are some odd references because of it. 

Maybe try two steps to make our agent startup behaviour better organized, given `__init__.py` was never intended to be this big and also not supposed to have logic scattered around in other modules. 

1. Follow the existing style to see how you can write the MeterService class without the Threading part, like the existing profiling service (put the threading part back to `agent/__init__.py`. (If you understand this already, look at step 2)
2. Try to make the existing `__init__.py` into more organized modules at the `agent/` directory; essentially, minimize __init__.py and decouple the current mess. (If you meet obstacles, revert to step 1, and it will be good enough)

Superskyyy: > I believe `__` prefixed variables are not supposed to be referred to by external classes, but when it is defined outside of classes, it doesn't do much other than the meaning of "you probably shouldn't use (import) this".
> 
> Anyway, I see the current implementation bit out of sync with the previous reporter styles.
> 
> That is, you provided a new `MeterService(Thread)`, and the result of defining threads outside of `agent/__init__.py` is that the code base becomes weird in terms of calling topology, and there are some odd references because of it.
> 
> Maybe try two steps to make our agent startup behaviour better organized, given `__init__.py` was never intended to be this big and also not supposed to have logic scattered around in other modules.
> 
> 1. Follow the existing style to see how you can write the MeterService class without the Threading part, like the existing profiling service (put the threading part back to `agent/__init__.py`. (If you understand this already, look at step 2)
> 2. Try to make the existing `__init__.py` into more organized modules at the `agent/` directory; essentially, minimize `__init__.py` and decouple the current mess. (If you meet obstacles, revert to step 1, and it will be good enough)

wu-sheng:
> > Sorry for messing up the commit history.
> 
> It's okay, everything will be squashed in the end.

Let's clear this policy. The result is YES, we would squash all the commits to merge as one single commit in the main branch. 
But this is only a tradeoff, as many new contributors don't know how to make sure the commit logs clear and readable. 

A real good commit log should be like this

https://github.com/apache/skywalking/commit/8a8070025afc9647ec842f053edd4c770fef1c90

Introducing the clear context of why this PR changed in this way and how we apply the changes. Multiple commit logs could be automatically merged into one commit log in merging.

Superskyyy:
> A real good commit log should be like this
> 
> [apache/skywalking@8a80700](https://github.com/apache/skywalking/commit/8a8070025afc9647ec842f053edd4c770fef1c90)
> 
> Introducing the clear context of why this PR changed in this way and how we apply the changes. Multiple commit logs could be automatically merged into one commit log in merging.

Really good example, thanks!

jiang1997:
> > > Sorry for messing up the commit history.
> > 
> > 
> > It's okay, everything will be squashed in the end.
> 
> Let's clear this policy. The result is YES, we would squash all the commits to merge as one single commit in the main branch. But this is only a tradeoff, as many new contributors don't know how to make sure the commit logs clear and readable.
> 
> A real good commit log should be like this
> 
> [apache/skywalking@8a80700](https://github.com/apache/skywalking/commit/8a8070025afc9647ec842f053edd4c770fef1c90)
> 
> Introducing the clear context of why this PR changed in this way and how we apply the changes. Multiple commit logs could be automatically merged into one commit log in merging.

Thanks for elaborating!

skywalking/agent/__init__.py
jiang1997: Got it.

jiang1997:
It's kind of weird. I passed ``make lint`` in my local environment but still failed in Actions.
Here is the message from ``CI / License and Lint (pull_request)``
```
venv/bin/flake8 .
./tests/plugin/base.py:35:1: B024 TestPluginBase is an abstract base class, but it has no abstract methods. Remember to use @abstractmethod, @abstractclassmethod and/or @abstractproperty decorators.
class TestPluginBase(ABC):
^
./skywalking/agent/protocol/__init__.py:22:1: B024 Protocol is an abstract base class, but it has no abstract methods. Remember to use @abstractmethod, @abstractclassmethod and/or @abstractproperty decorators.
class Protocol(ABC):
^
./skywalking/trace/span.py:36:1: B024 Span is an abstract base class, but it has no abstract methods. Remember to use @abstractmethod, @abstractclassmethod and/or @abstractproperty decorators.
class Span(ABC):
^
3     B024 Protocol is an abstract base class, but it has no abstract methods. Remember to use @abstractmethod, @abstractclassmethod and/or @abstractproperty decorators.
3
make: *** [Makefile:46: lint] Error 1
Error: Process completed with exit code 2.
```

Superskyyy:
> It's kind of weird. I passed `make lint` in my local environment but still failed in Actions. Here is the message from `CI / License and Lint (pull_request)`

Indeed weird, but what the linter pointed out was right. I'm not sure what happened.

jiang1997:
> > It's kind of weird. I passed `make lint` in my local environment but still failed in Actions. Here is the message from `CI / License and Lint (pull_request)`
> 
> Indeed weird, but what the linter pointed out was right. I'm not sure what happened.

How about creating a new pr to fix these errors that linter reported? I'm willing to.

Superskyyy:
> 



> > > It's kind of weird. I passed `make lint` in my local environment but still failed in Actions. Here is the message from `CI / License and Lint (pull_request)`
> > 
> > 
> > Indeed weird, but what the linter pointed out was right. I'm not sure what happened.
> 
> How about creating a new pr to fix these errors that linter reported? I'm willing to.

Well technically you could, but the PR cannot be allowed to merge until CI passes, unless you can bypass it.

kezhenxu94:
@jiang1997 can you please fix the linter errors in another PR. This should be because we didn’t pin the linter version and the linter updated recently. I just tried rerun a previously successful CI and it now failed.  Anyway, the linter errors in this PR look reasonable to be fixed. Please fix them in another PR

jiang1997:
> Generally looks good; please add the PVM metrics in the following PR and provide additional E2E test cases (refer to the currently commented `TODO` in the e2e .yaml file)
> 
> @kezhenxu94 Please also take a look :)

Got it.

Superskyyy:
Please take a look at the dead link check.

wu-sheng:
I think you should update the docs about meter APIs. @Superskyyy We also miss the tracing APIs doc.

Superskyyy:
> I think you should update the docs about meter APIs. @Superskyyy We also miss the tracing APIs doc.

Ok, we will provide both API docs next!

wu-sheng:
> reports roughly every one second.

I noticed this, is it configurable? Usually metrics don't report so frequently considering backend load and necessarily.
10-20s period is a better default value

jiang1997:
> > reports roughly every one second.
> 
> I noticed this, is it configurable? Usually metrics don't report so frequently considering backend load and necessarily. 10-20s period is a better default value

Thanks for the suggestion. I will make it configurable later.

Superskyyy:
Thanks!

docs/en/setup/advanced/MeterReporter.md
```diff
@@ -0,0 +1,110 @@
+# Python Agent Meter Reporter
```
kezhenxu94, line 1: Can you please also add a link in the menu to this doc?

https://github.com/apache/skywalking-python/blob/aad9034d4898a7f505f8347f28504353138a59d9/docs/menu.yml#L34-L38

docs/en/setup/advanced/MeterReporter.md
```diff
@@ -0,0 +1,110 @@
...
+os.environ['SW_AGENT_METER_REPORTER_ACTIVE'] = 'False'
+``` 
+or
+```bash
+export SW_AGENT_METER_REPORTER_ACTIVE=False
+```
+## Counter
+* `Counter` API represents a single monotonically increasing counter, automatic collect data and report to backend.
+```python
+tg_ls = [MeterTag("key", "value")]
```
kezhenxu94, line 26: Would be awesome if we can just pass in a Python tuple

```suggestion
tg_ls = [("key", "value")]
```

docs/en/setup/advanced/MeterReporter.md
```diff
@@ -0,0 +1,110 @@
...
+c = Counter('c2', CounterMode.INCREMENT, tg_ls)
+# or with more compact way
+# Counter('c2', CounterMode.INCREMENT).tag('key1', 'value1').tag('key2', 'value2')
+c.build()
+c.increment(2)
+```
+### Syntactic sugars
+```python
+c = Counter('c2', CounterMode.INCREMENT)
+c.build()
```
kezhenxu94, line 36: Looks like every meter should invoke `build` after construction, what about just embed the `build` inside the construction of the meters? So that users won't forget to call the `build` method?

kezhenxu94 (approved):
Hi @jiang1997, thanks for the patience and hard work!! I'm going to merge this PR to unblock your next steps, also I left some comments that you might want address in a following minor PR, or in your next large PR with the PVM metrics.

docs/en/setup/advanced/MeterReporter.md
jiang1997: Sounds like a good idea.

docs/en/setup/advanced/MeterReporter.md
jiang1997: Got it.

docs/en/setup/advanced/MeterReporter.md
jiang1997: You are right. This is the sequelae of imitating java agent.

wu-sheng: The reason we need build method because we separate tags definition from construction.

jiang1997: > The reason we need build method because we separate tags definition from construction.

Thanks for reminding. Here also separate tags definition from construction, so seems `build` is still needed.

skywalking/client/__init__.py
March225: @jiang1997 I collected the call times of an interface through Counter and reported it to skywalking. Now I want to display this metric on the ui page. How do I do it? Thank you!

wu-sheng: Learn MAL and Dashboard setup.
- https://skywalking.apache.org/docs/main/next/en/setup/backend/backend-meter/
- https://skywalking.apache.org/docs/main/next/en/concepts-and-designs/mal/
- Examples, https://github.com/apache/skywalking/tree/89d14e6256338840cd4a9b37f76980c8558e50c3/oap-server/server-starter/src/main/resources/meter-analyzer-config/
- Dashboard, https://skywalking.apache.org/docs/main/next/en/ui/readme/

jiang1997: @March225
This is just a simple supplement to wu-sheng's advice for the sake of simplicity.
1. Try to append your metric's definition to the server-side [config](https://github.com/apache/skywalking/blob/master/oap-server/server-starter/src/main/resources/meter-analyzer-config/python-runtime.yaml#L31-L36) file like below
https://github.com/apache/skywalking/blob/e2f1c0eadb52082e16f3abfc29dd8860000b596b/oap-server/server-starter/src/main/resources/meter-analyzer-config/python-runtime.yaml#L31-L36
Then edit the dashboard and add new widget to show your metric.
2. If nothing goes wrong, you should find your metric here.
image

PR #232, opened 2022-08-28
style: deal with B024 added to flake8 recently

jiang1997:
More details about B024 here https://github.com/PyCQA/flake8-bugbear/pull/274

## ./skywalking/agent/protocol/__init__.py:22:1: B024 Protocol is an abstract base class, but it has no abstract methods.
Set 'heartbeat', 'report' and 'report_log' as abstract method

## ./skywalking/trace/span.py:36:1: B024 Span is an abstract base class, but it has no abstract methods.
~Prevent base class Span from being directly instantiated by raising error in ``__new__`` with reference to
https://stackoverflow.com/questions/7989042/preventing-a-class-from-direct-instantiation-in-python~

## ./tests/plugin/base.py:35:1: B024 TestPluginBase is an abstract base class
Due to pytest instantiating parent class ~for searching test functions~, TestPluginBase has to be instantiable.

### DIFF
--- a/skywalking/trace/span.py
+++ b/skywalking/trace/span.py
@@ -16,7 +16,6 @@
 #
 
 import time
-from abc import ABC
 from collections import defaultdict
 from typing import List, Union, DefaultDict
 from typing import TYPE_CHECKING
@@ -33,7 +32,12 @@ if TYPE_CHECKING:
 
 
 @tostring
-class Span(ABC):
+class Span:
+    def __new__(cls, *args, **kwargs):
+        if cls is Span:
+            raise TypeError(f"only children of '{cls.__name__}' may be instantiated")
+        return super().__new__(cls, *args, **kwargs)
+
     def __init__(
             self,
             context: 'SpanContext',
--- a/skywalking/trace/span.py
+++ b/skywalking/trace/span.py
@@ -33,11 +33,6 @@ if TYPE_CHECKING:
 
 @tostring
 class Span:
-    def __new__(cls, *args, **kwargs):
-        if cls is Span:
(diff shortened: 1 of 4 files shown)


skywalking/trace/span.py
```diff
@@ -33,7 +32,24 @@
...
+            peer: str = None,
+            kind: Kind = None,
+            component: Component = None,
+            layer: Layer = None,
+            *args,
+            **kwargs,
+    ):
+        if cls is Span:
+            raise TypeError(f"only children of '{cls.__name__}' may be instantiated")
+        return super.__new__(cls, *args, **kwargs)
```
Superskyyy, line 51: `super()`?

jiang1997: Thanks!

jiang1997:
According to this line,
https://github.com/apache/skywalking-python/blob/5aca0dfb81a099f904647a81c570380ffacd3e52/skywalking/trace/context.py#L122
`Span` has to be instantiable too. And seems no alternatives for `Span`.

Superskyyy:
> So far, the only problem is that the test of flask uses `@runnable`, and according to this line,
> 
> https://github.com/apache/skywalking-python/blob/5aca0dfb81a099f904647a81c570380ffacd3e52/skywalking/trace/context.py#L122
> 
> 
> `Span` has to be instantiable too.
> Seems to be no alternatives for `Span`. Should I keep `Span` unchanged except removing `ABC`?

Yeah just keep it for now. :)

Superskyyy:
LGTM

Superskyyy (approved):

PR #236, opened 2022-09-13
refactor: hide some functions related to meter from users

jiang1997:
Introduce class `Builder` to instantiate classes derived from `Meter` to hide internal details from users.

@kezhenxu94 reminds me that `BaseMeter.tag()` is only allowed to be called arbitrary times before build and should not be called anymore after build. So it might be more reasonable to do so.


Superskyyy:
Dude, it's 3 am; go sleep!

kezhenxu94 (approved):

PR #245, opened 2022-10-24
Greenlet

westarest:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,7 +18,6 @@
 #
 .idea/
 *.iml
-.env/
 .DS_Store
 *~
 **/*_pb2*
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@
 #
 .idea/
 *.iml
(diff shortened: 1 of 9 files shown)


kezhenxu94 (changes requested):
This doesn't look to be ready for review

- Documentations should be in English.
- Some random files `main.py`, `gunicorn.conf.py`, etc., are put in the root directory without any references, if these are for development tests, remove them, if they are for testing, add tests to use them.
- Commented codes should be removed, don't use comments to backup codes

wu-sheng:
**Documentations should be in English.**

Besides this, the documents are in https://github.com/apache/skywalking-python/tree/master/docs and indexed by `menu.yaml`, hosted here, https://skywalking.apache.org/docs/skywalking-python/next/readme/

Test should be done through e2e, rather than a test main file.

westarest:
sorry ,this a draft, I only want to PR to my own Project for more test

PR #251, opened 2022-11-14
Enable E2E batch mode and pin Infra-E2E version

Superskyyy:
Batch E2E outputs to avoid excessive log.


Superskyyy:
This PR is no longer needed right? Since it's enabled by default now.

wu-sheng:
Do the project lock the version?

kezhenxu94:
> This PR is no longer needed right? Since it's enabled by default now.

Yes, but I suggest we should pin the github action to a specific revision/commit

Superskyyy:
Ok let me lock it (was previously curious to see if we could spot e2e bugs by applying @main here )

wu-sheng (approved):

PR #255, opened 2022-11-24
change the generator of traceId to the 3-parts

wenbindu:
update the traceid format.


wu-sheng:
Why are you proposing this? This seems never to be discussed and have conclusion.

Superskyyy:
The code doesn't seem reasonable and it's a core change. Please elaborate the purpose/benefit of this?

PR #260, opened 2022-12-15
support header namespace isolation

pingliu:
[] support header namespace isolation


wu-sheng:
This had been removed from the protocol months ago. We don't support this anymore.

Ref https://github.com/apache/skywalking-java/pull/123

Change log
- [Breaking Change] Remove the namespace from cross process propagation key.

The points we did are
1. **This is never defined in the protocol/APIs.**
1. Breaking a distributed tracing seems unreasonable because they are really accessing each other.
1. In four years of history, there is no other agent tried to support this, and we think <1> and few use cases are the reasons.

pingliu:
ok

PR #263, opened 2022-12-29
Enhance redis plugin to adapt Virtual Cache

Jedore:
Resolves: [#10212](https://github.com/apache/skywalking/issues/10212)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
   - Add reporter for PVM runtime metrics (default:disabled) (#238, #247)
   - Add Greenlet profiler (#246)
   - Add test and support for Python Slim base images (#249)
+  - Add support for the tags of Virtual Cache for Redis (#10212)
 
 - Plugins:
   - Add aioredis, aiormq, amqp, asyncpg, aio-pika, kombu RMQ plugins (#230 Missing test coverage) 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
   - Add reporter for PVM runtime metrics (default:disabled) (#238, #247)
   - Add Greenlet profiler (#246)
   - Add test and support for Python Slim base images (#249)
-  - Add support for the tags of Virtual Cache for Redis (#10212)
+  - Add support for the tags of Virtual Cache for Redis (#263)
 
 - Plugins:
   - Add aioredis, aiormq, amqp, asyncpg, aio-pika, kombu RMQ plugins (#230 Missing test coverage) 
--- a/skywalking/plugins/sw_redis.py
+++ b/skywalking/plugins/sw_redis.py
@@ -86,7 +86,7 @@ OPERATIONS_WRITE = set({'GETSET',
                         'XADD',
                         'XDEL',
                         'DEL',
-                        'xtrim'})
+                        'XTRIM'})
 
 OPERATIONS_READ = set({'GETRANGE',
                        'GETBIT ',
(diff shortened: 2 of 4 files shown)


Jedore:
I'just a little confused. Do I have to make these checks all successful ?

Superskyyy:
Hello @Jedore, yes you will have to make all checks pass to be eligible for a merge. The failed tests indicate problems in redis==3.5 plugin test, you can see the log in actions log.

I'm recovering from Covid now so I don't check github very often, please ping me here when it passes, I will take a review then.

Jedore:
All checks have passed. @Superskyyy

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
   - Add reporter for PVM runtime metrics (default:disabled) (#238, #247)
   - Add Greenlet profiler (#246)
   - Add test and support for Python Slim base images (#249)
+  - Add support for the tags of Virtual Cache for Redis (#10212)
```
Superskyyy, line 11: we use the # of PR here I think, so should be (#263)

```suggestion
  - Add support for the tags of Virtual Cache for Redis (#263)
```

skywalking/plugins/sw_redis.py
```diff
@@ -35,16 +35,121 @@ def install():
...
+            span.tag(TagCacheKey(key))
+            span.tag(TagCacheCmd(cmd))
+            span.tag(TagCacheOp(op))
 
             return res
 
     Connection.send_command = _sw_send_command
+
+
+OPERATIONS_WRITE = set({'GETSET',
```
Superskyyy, line 62: Is there a place where you get this complete set of supported commands for redis-py? I mean in case there are updates to supported commands we need a reference list to update the two sets. (if you have a url please simply add a comment to the url)

Superskyyy: Btw please move the definition of globals to top to reflect a Pythonic convention.

Superskyyy (changes requested):
Generally good, just some minor things. Thank you.

skywalking/plugins/sw_redis.py
Jedore: > Is there a place where you get this complete set of supported commands for redis-py? I mean in case there are updates to supported commands we need a reference list to update the two sets. (if you have a url please simply add a comment to the url)

The commands in https://github.com/apache/skywalking-java/blob/main/apm-sniffer/apm-sdk-plugin/jedis-plugins/jedis-4.x-plugin/src/main/java/org/apache/skywalking/apm/plugin/jedis/v4/JedisPluginConfig.java#L52  by [#10211](https://github.com/apache/skywalking/discussions/10211)

skywalking/plugins/sw_redis.py
```diff
@@ -27,6 +27,101 @@
...
+                        'ZREMRANGEBYRANK',
+                        'ZREMRANGEBYSCORE',
+                        'ZREVRANGE',
+                        'ZREVRANGEBYSCORE',
+                        'ZREVRANK',
+                        'ZUNIONSTORE',
+                        'XADD',
+                        'XDEL',
+                        'DEL',
+                        'xtrim'})
```
pg-yang, line 89: `xtrim`'s name style is different from others. Please confirm it.

Jedore: Thanks, I have confirmed it.

Superskyyy (approved):

Superskyyy:
LGTM thank you for the contribution!

PR #265, opened 2023-01-09
Fix dev dependency version constraint for Apple M1

Superskyyy:
Someone mentioned cannot install psycopg[binary], bumping version up will fix it for M1 macs.


wu-sheng:
Do you have m1 to verify? Do you need my local verification?

Superskyyy:
> Do you have m1 to verify? Do you need my local verification?

I don't have one, if it's convinient please help verify it by running `make env`, should be enough to see if the updated dependency installation works.

Superskyyy:
I talked with the dude from QQ group and he said this change works for his M2 mac. I think we are all good here.

PR #268, opened 2023-01-17
Use slim in plugin tests

Superskyyy:
Use slim to avoid long artifact build/download time in actions. (image size reduced by half) 

If I understand correctly slim is only different in terms of minimal external-to-python libs and should in no way interfere with plugin tests. (build-essential is additionally installed in image)

Please indicate if there's potential risk in this one.


Superskyyy (commented):
Something is wrong with the tests, investigating please don't merge

kezhenxu94 (dismissed):
LGTM when you mark it ready for review and CI passed 🧐

Superskyyy:
> LGTM when you mark it ready for review and CI passed 🧐

It will likely stuck forever 😂 Just found out there was a missed system-level dependency (for mysql-client) in the image.

Superskyyy:
I also updated compose retries and ci output logic. It will never hang for 30minutes anymore.

kezhenxu94 (dismissed):

Superskyyy:
zzZ my bad again.

wu-sheng (approved):

Superskyyy:
Finially😔

wu-sheng (approved):

PR #271, opened 2023-01-20
Add tables to show agent features and clean up configs

Superskyyy:
As Python agent grows and syncing features with the Java agent, even myself don't remember what was supported especially we have 3 reporter protocols (grpc http kafka) that not all of them support all telemetry data. 

So adding two tables to show it clearly for our users/developers so they can quick start and avoid installing the wrong version.

Preview:
image


wu-sheng:
I can't find meter and log docs, but the table shows supported.
Are those missed docs?

Superskyyy:
> I can't find meter and log docs, but the table shows supported. Are those missed docs?

The log reporter doc is here, but meter is missing. It's because we didn't add it to the menu.yaml, adding now.
image

Superskyyy:
@wu-sheng Btw, I remember HTTP handler (receiver) on meter is not implemented on the OAP side, is it intentional or just a missed thing not done?

wu-sheng:
It is just never defined from API level, and not implemented.
Feel free to add the API and implement it.

wu-sheng (approved):

PR #273, opened 2023-01-30
Remove configuration debts before we release 1.0.0

Superskyyy:
A considerable number of options are not well-formated in terms of their environment variable name. which brings headache to both maintainance and users. 

**This PR introduces breaking changes since configurations are renamed, this is explicitely stated at top of v1.0.0 changelog.**

Now I also added a documentation generator and checker to enforce a good and fully automatic way to produce configuration docuementation on the website. 

Changes in this PR:
1. Renamed a bunch of configs mainly envvar names to follow `SW_AGENT_`, also prioritized important configs.
2. Added doc generator, config-envvar sync checker, ci config checker
3. Fix a potential undetected problem in profiler code

I've carefully reviewed all changes, hopefully nothing goes wrong.


Superskyyy:
Fixing a deadlink.

wu-sheng (approved):

tests/plugin/docker-compose.base.yml
```diff
@@ -35,10 +35,9 @@ services:
   agent:
     image: apache/skywalking-python-agent:latest-plugin
     environment:
-      SW_AGENT_COLLECTOR_BACKEND_SERVICES: collector:19876
```
kezhenxu94, line 38: Some of the env var names (like this one) were deliberately given that names to keep it consistent with other agent's configuration, although I agree you rename to a better one.

Superskyyy: oO I didn't see this thread until now, do you think we should tweak some naming to sync with java agent?

Superskyyy: > oO I didn't see this thread until now, do you think we should tweak some naming to sync with java agent?

From the point of python keyed parameter feature, it maybe inconvinient to have a long name, so I generally shortened them a bit. Anyway I will think about it and update some if needed. Next version I suppose we will provide an actual config.toml file instead of fully relying on env vars or code changes.

kezhenxu94: > > oO I didn't see this thread until now, do you think we should tweak some naming to sync with java agent?
> 
> 
> 
> From the point of python keyed parameter feature, it maybe inconvinient to have a long name, so I generally shortened them a bit. 

Can we shorten the parameter name in code while keep the original env var name? I think the parameter name doesn't have to be the same as env var name right?

> Anyway I will think about it and update some if needed. Next version I suppose we will provide an actual config.toml file instead of fully relying on env vars or code changes.

We can have a toml file for configuration for sure. But all the configurations in that file should be overridable via env var, that would be convenient in Docker/Kubernetes use case.

Superskyyy: > 

Yeah it doens't have to be the same just I think it's a convention to have that consistent env_prefix_var as I seen in other python libraries. I will make some changes tmr to make important envvars back in sync with Java agent, meanwhile I think it's okay to sacrafice the length consideration in exchange for consistency (to be the same as envvar).

PR #276, opened 2023-02-05
Add loguru support

Jedore:
### Add log reporter support for loguru
- [ ] If adding a new plugin, add a component id in [the main repo](https://github.com/apache/skywalking/blob/master/oap-server/server-starter/src/main/resources/component-libraries.yml)
- [ ]  If adding a new plugin, add a logo in [the UI repo](https://github.com/apache/skywalking-booster-ui/tree/main/src/assets/img/technologies)
- [x] I have added the library to `pyproject.toml` (plugin group) by running `poetry add library --group plugins`
- [x] I have rebuilt the `Plugins.md` documentation by running `make doc-gen`

- [x] Closes: https://github.com/apache/skywalking/issues/10334


tests/plugin/data/sw_loguru/expected.data.yml
```diff
@@ -0,0 +1,61 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+logItems:
+  - serviceName: provider
+    logSize: 2
```
wu-sheng, line 20: Is a very precise log size good? Would this test generate more than 2 records? Such as retry.

wu-sheng:
In the issue discussion, there is a topic about `python native logger and loguru co-exist`. What will happen in this case?

Jedore:
I would supply tomorrow.

Superskyyy:
You forgot to install fastapi in the docker-compose command. That's why ci failed. Other than this looks good.

tests/plugin/data/sw_loguru/expected.data.yml
Jedore: > Is a very precise log size good? Would this test generate more than 2 records? Such as retry.

Got it.

kezhenxu94 (approved):
Looks good, thanks. Fix the CI and let's merge!

Jedore:
> Looks good, thanks. Fix the CI and let's merge!

Thank you.

Jedore:
**loguru and logging are different log system，they don't affect each other. Like in `tests/plugin/sw_loguru/services/provider.py`,  loguru generates 2 logitems, logging genereate 2 logitems.**

wu-sheng:
> And i'm not familiar with yaml. expected.data.yml takes too much time and still seems not valid. 😢

What do you mean? Anything blocks you?

Jedore:
> > And i'm not familiar with yaml. expected.data.yml takes too much time and still seems not valid. 😢
> 
> What do you mean? Anything blocks you?

I run the test command in my repo, but failed, as:
```text
=================================== FAILURES ===================================
____________________ TestPlugin.test_plugin[loguru==0.6.0] _____________________

self = <tests.plugin.data.sw_loguru.test_loguru.TestPlugin object at 0x7fae8eb14fd0>
docker_compose = <testcontainers.compose.DockerCompose object at 0x7fae8ea683a0>
version = 'loguru==0.6.0'

    @pytest.mark.parametrize('version', get_test_vector(lib_name='loguru', support_matrix=support_matrix))
    def test_plugin(self, docker_compose, version):
>       self.validate()

tests/plugin/data/sw_loguru/test_loguru.py:36: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <tests.plugin.data.sw_loguru.test_loguru.TestPlugin object at 0x7fae8eb14fd0>
expected_file_name = '/home/runner/work/skywalking-python/skywalking-python/tests/plugin/data/sw_loguru/expected.data.yml'

    def validate(self, expected_file_name=None):
        # type: (str) -> Response
    
        if expected_file_name is None:
            expected_file_name = os.path.join(dirname(inspect.getfile(self.__class__)), 'expected.data.yml')
    
        # time.sleep(10)
    
        with open(expected_file_name) as expected_data_file:
            expected_data = os.linesep.join(expected_data_file.readlines())
    
            response = requests.post(url='http://localhost:12800/dataValidate', data=expected_data)
    
            if response.status_code != 200:
                res = requests.get('http://localhost:12800/receiveData')
    
                actual_data = yaml.dump(yaml.load(res.content, Loader=Loader))
    
                differ = Differ()
                diff_list = list(differ.compare(
                    actual_data.splitlines(keepends=True),
                    yaml.dump(yaml.load(expected_data, Loader=Loader)).splitlines(keepends=True)
                ))
    
                print('diff list: ')
    
                sys.stdout.writelines(diff_list)
>           assert response.status_code == 200
E           AssertionError

tests/plugin/base.py:62: AssertionError
---------------------------- Captured stdout setup -----------------------------
7f20775e0631356c4823d9372b09d653db0e6540: Pulling from apache/skywalking-agent-test-tool/mock-collector
Digest: sha256:8c010883c8203a34f1239c6487855c5ba5b40a4c52faf8d21ccc3b5073438[221](https://github.com/Jedore/skywalking-python/actions/runs/4102318884/jobs/7075166151#step:6:222)
Status: Downloaded newer image for ghcr.io/apache/skywalking-agent-test-tool/mock-collector:7f20775e0631356c4823d9372b09d653db0e6540
---------------------------- Captured stderr setup -----------------------------
Creating network "sw_loguru_beyond" with the default driver
Pulling collector (ghcr.io/apache/skywalking-agent-test-tool/mock-collector:7f20775e0631356c4823d9372b09d653db0e6540)...
Creating sw_loguru_collector_1 ... 
Creating sw_loguru_collector_1 ... done
Creating sw_loguru_provider_1  ... 
Creating sw_loguru_provider_1  ... done
Creating sw_loguru_consumer_1  ... 
Creating sw_loguru_consumer_1  ... done
----------------------------- Captured stdout call -----------------------------
diff list: 
  logItems:
  - logSize: 4
    logs:
    - body:
        content:
          text: Loguru provider error reported.
        type: text
      endpoint: /users
      layer: ''
      tags:
        data:
        - key: level
          value: ERROR
        - key: logger
          value: loguru
        - key: thread
-         value: MainThread
+         value: not null
        - key: exception
+         value: not null
+     timestamp: gt 0
-         value: "Traceback (most recent call last):\n  File \"/app/services/provider.py\"\
-           , line 36, in application\n    raise Exception('Loguru Exception Test.')\n\
-           Exception: Loguru Exception Test.\n"
-     timestamp: 1675678728625
      traceContext:
        spanId: 0
-       traceId: a4f8cfd2a60711edb5500[242](https://github.com/Jedore/skywalking-python/actions/runs/4102318884/jobs/7075166151#step:6:243)ac120004
-       traceSegmentId: a4f95e52a60711edb8a30242ac120003
+       traceId: not null
+       traceSegmentId: not null
    - body:
        content:
+         text: not null
-         text: "2023-02-06 10:18:49,125 [MainThread] ERROR root - Logging provider\
-           \ error reported.\nTraceback (most recent call last):\n  File \"/app/services/provider.py\"\
-           , line 36, in application\n    raise Exception('Loguru Exception Test.')\n\
-           Exception: Loguru Exception Test."
        type: text
      endpoint: /users
      layer: ''
      tags:
        data:
        - key: level
          value: ERROR
        - key: logger
          value: root
        - key: thread
-         value: MainThread
-     timestamp: 1675678729126
+         value: not null
+       - key: exception
+         value: not null
+     timestamp: gt 0
      traceContext:
        spanId: 0
-       traceId: a4f8cfd2a60711edb5500242ac120004
-       traceSegmentId: a4f95e52a60711edb8a30242ac120003
+       traceId: not null
+       traceSegmentId: not null
    - body:
        content:
          text: Loguru provider warning reported.
        type: text
      endpoint: /users
      layer: ''
      tags:
        data:
        - key: level
          value: WARNING
        - key: logger
          value: loguru
        - key: thread
-         value: MainThread
-     timestamp: 1675678730128
+         value: not null
+     timestamp: gt 0
      traceContext:
        spanId: 0
-       traceId: a4f8cfd2a60711edb5500242ac120004
-       traceSegmentId: a4f95e52a60711edb8a30242ac120003
+       traceId: not null
+       traceSegmentId: not null
    - body:
        content:
+         text: Logging provider info reported.
-         text: 2023-02-06 10:18:50,628 [MainThread] CRITICAL root - Logging provider
-           critical reported.
        type: text
      endpoint: /users
      layer: ''
      tags:
        data:
        - key: level
          value: CRITICAL
        - key: logger
          value: root
        - key: thread
-         value: MainThread
-     timestamp: 1675678730629
+         value: not null
+     timestamp: gt 0
      traceContext:
        spanId: 0
-       traceId: a4f8cfd2a60711edb5500242ac120004
-       traceSegmentId: a4f95e52a60711edb8a30242ac120003
+       traceId: not null
+       traceSegmentId: not null
    serviceName: provider
  meterItems: []
  segmentItems:
  - segmentSize: 1
    segments:
-   - segmentId: a4f95e52a60711edb8a30242ac120003
+   - segmentId: not null
      spans:
      - componentId: 7014
-       endTime: 1675678730630
+       endTime: gt 0
        isError: false
        operationName: /users
        parentSpanId: -1
-       peer: 172.18.0.4:48448
+       peer: not null
        refs:
-       - networkAddress: provider:9091
+       - networkAddress: not null
          parentEndpoint: /users
          parentService: consumer
-         parentServiceInstance: a124c208a60711edb5500242ac120004
+         parentServiceInstance: not null
          parentSpanId: 1
-         parentTraceSegmentId: a4f8ce4ca60711edb5500242ac120004
+         parentTraceSegmentId: not null,
          refType: CrossProcess
-         traceId: a4f8cfd2a60711edb5500242ac120004
+         traceId: not null
        skipAnalysis: false
        spanId: 0
        spanLayer: Http
        spanType: Entry
-       startTime: 16756787[281](https://github.com/Jedore/skywalking-python/actions/runs/4102318884/jobs/7075166151#step:6:282)14
+       startTime: gt 0
        tags:
        - key: http.method
          value: GET
        - key: http.url
-         value: http://provider:9091/users
+         value: not null
        - key: http.status_code
          value: '200'
    serviceName: provider
  - segmentSize: 1
    segments:
-   - segmentId: a4f8ce4ca60711edb5500242ac120004
+   - segmentId: not null
      spans:
      - componentId: 7002
-       endTime: 16756787[306](https://github.com/Jedore/skywalking-python/actions/runs/4102318884/jobs/7075166151#step:6:307)30
+       endTime: gt 0
        isError: false
        operationName: /users
        parentSpanId: 0
-       peer: provider:9091
+       peer: ''
        skipAnalysis: false
        spanId: 1
        spanLayer: Http
        spanType: Exit
-       startTime: 1675678728110
+       startTime: gt 0
+       tags:
+       - key: http.url
+         value: not null
+       - key: http.method
+         value: GET
+       - key: http.status_code
+         value: '200'
+   - segmentId: not null
+     spans:
+     - componentId: 7014
+       endTime: gt 0
+       isError: false
+       operationName: /users
+       parentSpanId: -1
+       peer: not null
+       skipAnalysis: false
+       spanId: 0
+       spanLayer: Http
+       spanType: Entry
+       startTime: gt 0
        tags:
        - key: http.method
          value: GET
        - key: http.url
+         value: not null
-         value: http://provider:9091/users
-       - key: http.status_code
-         value: '200'
-     - componentId: 7014
-       endTime: 1675678730636
-       isError: false
-       operationName: /users
-       parentSpanId: -1
-       peer: 172.18.0.1:50156
-       skipAnalysis: false
-       spanId: 0
-       spanLayer: Http
-       spanType: Entry
-       startTime: 1675678728110
-       tags:
-       - key: http.method
-         value: GET
-       - key: http.url
-         value: http://0.0.0.0:9090/users
        - key: http.status_code
          value: '200'
    serviceName: consumer
--------------------------- Captured stderr teardown ---------------------------
```

So i only can commited the fixed code firstly and look for your help.

wu-sheng:
+ means what is in the actual data
- means this is in your expect data, but can't find in the actual data.

Jedore:
> ```
> + means what is in the actual data
> - means this is in your expect data, but can't find in the actual data.
> ```

So actual data and expect data must be the same absoublutly?  And seems `not null / gt 0` .etc syntax is invalid ?

tests/plugin/data/sw_loguru/services/provider.py
```diff
@@ -24,21 +25,35 @@
...
 
+        # this will be filtered by SW_AGENT_LOG_REPORTER_LEVEL
         logger.debug('Loguru provider debug reported.')
 
+        time.sleep(0.5)
+
         logger.warning('Loguru provider warning reported.')
 
+        time.sleep(0.5)
+
```
kezhenxu94, line 52: Can we remove these `time.sleep`?

Jedore: > Can we remove these `time.sleep`?

Of course. I referred other test file.

Jedore:
It seems `not null / gt 0` .etc syntax is valid in other pluginj's expected.data.yml, but invalid in loguru's expected.data.yml. I'm confulsed and blocked. 🙃 
Can you give me some suggestions? ❤️

wu-sheng:
> It seems `not null / gt 0` .etc syntax is valid in other plugin's expected.data.yml, but invalid in loguru's expected.data.yml. I'm confulsed and blocked. 🙃 Can you give me some suggestions? ❤️

@jmjoy @mrproliu  Could you take a look? The log verification was added through https://github.com/apache/skywalking-agent-test-tool/pull/39

tests/plugin/data/sw_loguru/expected.data.yml
```diff
@@ -0,0 +1,185 @@
...
+          - operationName: '/users'
+            parentSpanId: 0
+            spanId: 1
+            spanLayer: Http
+            startTime: gt 0
+            endTime: gt 0
+            componentId: 7002
+            isError: false
+            spanType: Exit
+            peer: ''
```
kezhenxu94, line 156: This should not be empty, no?

kezhenxu94 (commented):
> It seems `not null / gt 0` .etc syntax is valid in other plugin's expected.data.yml, but invalid in loguru's expected.data.yml. I'm confulsed and blocked. 🙃
> Can you give me some suggestions? ❤️


I think they worked, when printing the diff logs in the final result, we keep the raw expected expression `not null` as is, so when checking the result you can safely ignore those diff containing `not null`, `gt 0` ect., see my comment, which might be the cause of the CI failure

tests/plugin/data/sw_loguru/expected.data.yml
Jedore: > This should not be empty, no?

👌

tests/plugin/data/sw_loguru/services/provider.py
Superskyyy: I think the sleep was added because of this bug https://github.com/apache/skywalking/issues/8752 that introduced flaky tests in e2e, but it shouldn't be a problem for you in plugin tests, so can be safely removed.

Superskyyy:
I will take a look at the yamls.

Superskyyy:
image
Is this the cause?🙃

Superskyyy:
One more thing, I see that there's no use of SW_formatter meaning no support for config.layout right? Maybe please provide such additional information like "only built-in logging module supports this configuration blabla, lorugu doesn't support this configuration blabla"

skywalking/plugins/sw_loguru.py
```diff
@@ -0,0 +1,224 @@
...
+            core.patcher(log_record)
+
+        if patcher:
+            patcher(log_record)
+
+        return log_record
+
+    def _sw_log(self, level_id, static_level_no, from_decorator, options, message, args, kwargs):
+        _log(level_id, static_level_no, from_decorator, options, message, args, kwargs)
+        record = gen_record(self, level_id, static_level_no, from_decorator, options, message, args, kwargs)
```
Superskyyy, line 153: NoneType bug here, your gen_record function can return None if core handler missing (from `logger.remove()`), then it will error line 156

Superskyyy (dismissed):

skywalking/plugins/sw_loguru.py
Jedore: > NoneType bug here, your gen_record function can return None if core handler missing (from `logger.remove()`), then it will error line 156

yeah, so carefully 👍

Jedore:
> One more thing, I see that there's no use of SW_formatter meaning no support for `log_reporter_layout` right? Maybe please provide such additional information like "only built-in logging module supports this configuration blabla, lorugu doesn't support this configuration blabla"
> 
> Same goes to `log_reporter_formatted` which is the trigger of layout.

loguru's format and record are different from logging's and SW_formatter is for logging record. So i think just use loguru raw message. 

Provide such additional information is good. But may be it's not important now.  Python Agent Log Reporter aims at logging module.  loguru just use a few configuration in https://skywalking.apache.org/docs/skywalking-python/next/en/setup/configuration/#log-reporter-configurations.

Jedore:
> image Is this the cause?🙃

You are awesome! 🚀

Superskyyy:
Nope, failure persists...

Superskyyy:
> > One more thing, I see that there's no use of SW_formatter meaning no support for `log_reporter_layout` right? Maybe please provide such additional information like "only built-in logging module supports this configuration blabla, lorugu doesn't support this configuration blabla"
> 
> > 
> 
> > Same goes to `log_reporter_formatted` which is the trigger of layout.
> 
> 
> 
> loguru's format and record are different from logging's and SW_formatter is for logging record. So i think just use loguru raw message. 
> 
> 
> 
> Provide such additional information is good. But may be it's not important now.  Python Agent Log Reporter aims at logging module.  loguru just use a few configuration in https://skywalking.apache.org/docs/skywalking-python/next/en/setup/configuration/#log-reporter-configurations. 
> 
> 

I'm just thinking a situation where user thinks some config like layout also applies to loguru so they set it but see no effect. They may get confused, so adding a comment to the config will help reduce this risk.

Jedore:
Seems https://github.com/Jedore/skywalking-python/blob/test/tests/plugin/data/sw_loguru/expected.data.yml and https://github.com/Jedore/skywalking-python/blob/test/tests/plugin/data/sw_loguru/services/provider.py are valid.
A simple action : https://github.com/Jedore/skywalking-python/actions/runs/4107258862

Superskyyy:
> > > > One more thing, I see that there's no use of SW_formatter meaning no support for `log_reporter_layout` right? Maybe please provide such additional information like "only built-in logging module supports this configuration blabla, lorugu doesn't support this configuration blabla"
> 
> > > 
> 
> > > 
> 
> > > > 
> 
> > > 
> 
> > > 
> 
> > > > Same goes to `log_reporter_formatted` which is the trigger of layout.
> 
> > > 
> 
> > > 
> 
> > > loguru's format and record are different from logging's and SW_formatter is for logging record. So i think just use loguru raw message.
> 
> > > Provide such additional information is good. But may be it's not important now.  Python Agent Log Reporter aims at logging module.  loguru just use a few configuration in https://skywalking.apache.org/docs/skywalking-python/next/en/setup/configuration/#log-reporter-configurations.
> 
> > 
> 
> > I'm just thinking a situation where user thinks some config like layout also applies to loguru so they set it but see no effect. They may get confused, so adding a comment to the config will help reduce this risk.
> 
> 
> 
> Good, add where ?

Simply add to the corresponding comment in config.py and regenerate documentation by make
doc-gen.

Superskyyy:
So from what I can see the only difference is that you removed the `exception:value` pair and changed the not null to value. That is truely weird right. The exception tag is supposed to be there but not being there. 

> Seems [Jedore/skywalking-python@`test`/tests/plugin/data/sw_loguru/expected.data.yml](https://github.com/Jedore/skywalking-python/blob/test/tests/plugin/data/sw_loguru/expected.data.yml?rgh-link-date=2023-02-06T19%3A53%3A07Z) and [Jedore/skywalking-python@`test`/tests/plugin/data/sw_loguru/services/provider.py](https://github.com/Jedore/skywalking-python/blob/test/tests/plugin/data/sw_loguru/services/provider.py?rgh-link-date=2023-02-06T19%3A53%3A07Z) are valid. A simple action : [Jedore/skywalking-python/actions/runs/4107258862](https://github.com/Jedore/skywalking-python/actions/runs/4107258862) continue test it ...

Jedore:
ok. i try.

Jedore:
I tested and i think the block point is `exception:value` pair of logging and consumer segments pair order.

kezhenxu94 (approved):
@Superskyyy is off some time for a rest, I'm merging this

Jedore:
This is my comitting for a complete python plugin. Although met some trouble, thank you very much. 🤛  @wu-sheng @kezhenxu94 @Superskyyy

PR #278, opened 2023-02-06
Add support for fork and periodical property report

Superskyyy:
Add periodical reporter for instance properties. 

Also I think I got fork() support working in the easiest possible way. Closes: https://github.com/apache/skywalking/issues/7319
It is now exposed as an experimental option, no formal e2e tests yet, but I have extensively tested with manual work. 

Now gRPC, HTTP and Kafka protocols all support fork, Trace, Log, Meter, Profiling all seem to work okay after fork.

Changelist:
1. Periodical instance property report.
2. Fork support (os.fork())
3. Rewrote agent startup module, now it's class-based (to avoid all kinds of problems).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
   - Add Greenlet profiler (#246)
   - Add test and support for Python Slim base images (#249)
   - Add support for the tags of Virtual Cache for Redis (#263)
+  - Add **experimental** support for explicit os.fork(), restarts agent in new process (#278)
 
 - Plugins:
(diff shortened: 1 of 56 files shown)


Superskyyy:
There could still be some nits to fix, but I'm overwhelmed currently, so leaving for a day and I will check back.

Superskyyy:
I think the fork support accidentally fixed something that the tests assumed to be otherwise... investigating

Superskyyy:
This is not ready, closing for now.

PR #293, opened 2023-03-15
Don't init profile / meter damon threads when not active.

ZEALi:
should not init profile / meter damon threads if those configuration are not active.

### DIFF
--- a/skywalking/agent/__init__.py
+++ b/skywalking/agent/__init__.py
@@ -259,8 +259,10 @@ class SkyWalkingAgent(Singleton):
             import grpc.experimental.gevent as grpc_gevent
             grpc_gevent.init_gevent()
 
-        profile.init()
-        meter.init(force=True)  # force re-init after fork()
+        if config.agent_profile_active:
+            profile.init()
+        if config.agent_meter_reporter_active:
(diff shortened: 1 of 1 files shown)


Superskyyy:
Good catch thanks

Superskyyy:
Let's wait CI passes.

wu-sheng:
@Superskyyy Why the deadlink happens? I assume this is a main branch issue.

Superskyyy:
> @Superskyyy Why the deadlink happens? I assume this is a main branch issue.
It's the mit gpg site, I will remove it from doc.

Superskyyy:
Please do some additional changes to tests. They cannot pass due to profiling service not available>

In context.py  - profiling is assumed to be enabled, which is not true according to this change. https://github.com/apache/skywalking-python/blob/1803a71bf242b8a13474c51550118f51c420762a/skywalking/trace/context.py#L151
```        # start profiling if profile_context is set
        if self.profile_status is None:
            self.profile_status = profile.profile_task_execution_service.add_profiling(self,
                                                                                       self.segment.segment_id,
                                                                                       op)
```

PR #296, opened 2023-03-17
Don't install loguru when not active

ZEALi:
As a log reporter, it should not be activated when user don't want (agent.archive_log will fail). 

maybe sw_loguru should be treated as sw_logging which is installed before other normal plugins in agent.start(),  
or re-factor sw_loguru & sw_logging into a "log reporter" plugins instead of default "trace reporter" plugins?

### DIFF
--- a/skywalking/plugins/sw_loguru.py
+++ b/skywalking/plugins/sw_loguru.py
@@ -40,6 +40,11 @@ note = """"""
 
 
 def install():
+    
+    if not config.agent_log_reporter_active:
+        # DO NOT even try to monkey-patch if not active, avoid any performance losing.
+        return
+
     from loguru import logger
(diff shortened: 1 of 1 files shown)


Superskyyy:
Please fix lint. 

Regarding your question, I think I would prefer the second design as it is more clear and support future extensions. Would you like to proceed on the refactor?

Superskyyy:
Please remember to add a changelog for this fix.

ZEALi:
> 

Sure. I'll work on the sepreated hook for “log reporter” plugins.

Superskyyy:
Any updates? @ZEALi

Superskyyy:
Closing this in favor of a newer PR including this change https://github.com/apache/skywalking-python/pull/302/files

PR #300, opened 2023-04-19
Fix high CPU usage

FAWC438:
About high CPU usage bug discussed in [issues#10672](https://github.com/apache/skywalking/issues/10672), my final solution was to set the `init_wait` parameter in the function decorator to 0.02 (20 ms). In addition, I also made additional changes to make the thread delay more reasonable and closer to the implementation of Java agent.

For the segment/log reporting function, I added a return value of bool type, which is only used by the `@report_with_backoff` decorator to determine whether the queue is not empty. If the queue is not empty, the thread will report the data immediately without any waiting; otherwise, the thread will wait for 20ms to release the CPU usage.

For other types of reporting functions, since their return value is always None, the function decorator still operates according to the original logic

I think the above changes balance the two solutions proposed by [issues#10672](https://github.com/apache/skywalking/issues/10672) and make it more similar to the Java agent implementation

### Fix <bug description or bug issue link>
- [ ] Add a unit test to verify that the fix works.
- [x] Explain briefly why the bug exists and how to fix it.
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue url. Closes: [issues#10672](https://github.com/apache/skywalking/issues/10672)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
 
 - Fixes:
   - Fix sw8 loss when use aiohttp.(#299,issue#10669)
+  - Fix the bug with high cpu usage.(#300,issue#10672)
 
 ### 1.0.0
 
--- a/skywalking/agent/__init__.py
+++ b/skywalking/agent/__init__.py
@@ -314,7 +314,7 @@ class SkyWalkingAgent(Singleton):
 
     @report_with_backoff(reporter_name='segment', init_wait=0.02)
     def __report_segment(self) -> bool:
-        '''Returns True if the queue is not empty'''
+        """Returns True if the queue is not empty"""
         queue_not_empty_flag = not self.__segment_queue.empty()
         if queue_not_empty_flag:
             self.__protocol.report_segment(self.__segment_queue)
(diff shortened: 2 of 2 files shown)


Superskyyy:
LGTM! I will kick off a release tomorrow if no objections.

Superskyyy (approved):

Superskyyy:
@kezhenxu94 FYI, as you also encountered this problem in 1.0.0

wu-sheng:
Are you going to cherry pick patches to do 1.0.1?
I think there is no block for this.

Superskyyy:
Yes, since its a patch version I will cherry pick them to do 1.0.1. Gonna figure out how to do that properly.

wu-sheng:
You should create `1.0.x` branch based on 1.0.0 release commit. Then cherry pick this to that branch and upgrade version number for the release.
The main branch stays for 1.1.0 still.

Superskyyy:
> You should create `1.0.x` branch based on 1.0.0 release commit. Then cherry pick this to that branch and upgrade version number for the release. The main branch stays for 1.1.0 still.

I see, will do!

PR #311, opened 2023-06-02
Fix unexpected 'No active span' IllegalStateError

ZEALi:
The _skywalking.trace.context.SpanContext.active_span()_ method has been refactored in v1.0.0. Its equivalent method has been changed to _peek()_ . When _peek()_ returns None, the _active_span()_ method throws **IllegalStateError** exception directly. 

This change is not backward compatible, SOME plugins that rely on the returned None value of _active_span()_ WILL be broken ( sw_flask, sw_sanic, sw_tornado, sw_django, etc.).

For example, when running Flask by **gunicorn** which not monkey-patched as well as for sw_http_server plugin, calling _get_context().active_span_ in __sw_handle_exception()_ will raise an IllegalStateError exception directly and break the original logic of the container's _handle_exception()_ .

In v0.8.0, no exception was thrown, so the caller could check if entry_span is not None and perform further processing. 

Therefore, I suggest retaining the behavioral pattern of the _active_span()_ method in v0.8.0 , and providing parameters in the added _peek()_ method to support scenarios where exceptions need to be thrown (such as in sw_logging / sw_loguru).

### DIFF
--- a/skywalking/log/sw_logging.py
+++ b/skywalking/log/sw_logging.py
@@ -83,7 +83,7 @@ def install():
             # exceptions before any span is even created, just ignore these fields and
             # avoid appending 'no active span' traceback that could be confusing.
             # Or simply the log is generated outside any span context.
-            active_span_id = context.active_span.sid
+            active_span_id = context.peek(raise_if_none=True).sid
             primary_endpoint_name = context.primary_endpoint.get_name()
         except IllegalStateError:
             pass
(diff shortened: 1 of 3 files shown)


Superskyyy:
Good catch! Thanks for the fix.

Superskyyy (approved):

wu-sheng:
CI fails, please fix them first.

Superskyyy:
Probably needs a full rerun， the old image got deleted last time.

Superskyyy:
Kafka cases are flaky, not related to this PR or agent functionalities, will fix.

PR #312, opened 2023-06-14
Add Neo4j plugin.

Jedore:
### <Feature description>
- [x] If adding a new plugin, add a component id in [the main repo](https://github.com/apache/skywalking/blob/master/oap-server/server-starter/src/main/resources/component-libraries.yml). As https://github.com/apache/skywalking/pull/10937
- [ ]  If adding a new plugin, add a logo in [the UI repo](https://github.com/apache/skywalking-booster-ui/tree/main/src/assets/img/technologies)
- [x] I have added the library to `pyproject.toml` (plugin group) by running `poetry add neo4j --group plugins`
- [x] I have rebuilt the `Plugins.md` documentation by running `make doc-gen`

- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue url. Closes: https://github.com/apache/skywalking/issues/10898

Something about the plugin,
The major Neo4j DBMS versions are 5 and 4.4. The latest neo4j python driver version is `5.x.x` which support both Neo4j DBMS 5 and 4.4 . And python driver `4.4.x` differs much from `5.x.x` .So the neo4j plugin only support neo4j python driver `5.x.x`. 
Reference:
> "The latest 5.x driver supports connection to any Neo4j 5 and 4.4 instance, and will also be compatible with Neo4j 6." in https://neo4j.com/docs/python-manual/5/install/

Neo4j python driver `5.x.x` provide two method `Session.run()`&`Transaction.run()` to execute Cypher statement with both *sync* and *async* mode. `tests/plugin/data/sw_neo4j/services/provider.py` contains these four cases:
1. sync `Session.run()`
2. sync `Transaction.run()`
3. async `Session.run()`
4. async `Transaction.run()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 ## Change Logs
 
 ### 1.1.0
+- Plugins:
+  - Add neo4j plugin.(#312)
 
 ### 1.0.1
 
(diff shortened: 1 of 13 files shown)


Superskyyy:
Generally looks great! Please add the neo4j dependency to poetry plugins group as indicated in the PR template. And if you would like maybe also add a note in the neo4j plugin (You see there's a variable called note there), to indicate only the 5.x driver is supported and its fine since it can support both DBMS versions.

Jedore:
@Superskyyy Thanks great! Please review the note and prproject.toml.

Superskyyy (approved):

PR #323, opened 2023-10-11
Add support printing TID to logs

CodePrometheus:
- [x] If this pull request closes/resolves/fixes an existing issue, replace the issue url. Closes: [https://github.com/apache/skywalking/issues/10734](https://github.com/apache/skywalking/issues/10734)


wu-sheng:
Please update the docs accordingly.

CodePrometheus:
> Please update the docs accordingly.

done

wu-sheng:
@Superskyyy Is there any doc about introducing how to use the logger?

Superskyyy:
> @Superskyyy Is there any doc about introducing how to use the logger?

Not about this particular feature, @CodePrometheus would you mind adding a small doc to this [section ](https://skywalking.apache.org/docs/skywalking-python/next/en/setup/advanced/logreporter/#python-agent-log-reporter)(by adding a new subsection called something like ``printing traceID to logs".

skywalking/loggings.py
```diff
@@ -22,12 +22,21 @@
...
+        context = get_context()
+        record.tid = str(context.segment.related_traces[0])
+        return True
+
+
 def getLogger(name=None):  # noqa
     logger = logging.getLogger(name)
     ch = logging.StreamHandler()
-    formatter = logging.Formatter('%(asctime)s %(name)s [pid:%(process)d] [%(threadName)s] [%(levelname)s] %(message)s')
+    formatter = logging.Formatter('%(asctime)s %(name)s %(tid)s [pid:%(process)d] [%(threadName)s] [%(levelname)s] %(message)s')
```
Superskyyy, line 36: This feature should be optional as consistent with the Java agent (see [here](https://skywalking.apache.org/docs/skywalking-java/next/en/setup/service-agent/java-agent/application-toolkit-logback-1.x/#print-trace-id-in-your-logs)), because not all users want tids in their logs. To achieve this behavior in Python, please introduce a new setting to control the insertion of tids and adding the additional filter.

Superskyyy:
Oh I just realized your understanding of this feature from conversation in [#323](https://github.com/apache/skywalking/issues/10734#issuecomment-1752038813) was not entirely accurate. Sorry, I should have pointed out earlier when you were confirming it. The place you put this code was wrong, `skywalking/loggings.py` is the dedicated place to hold logs of **the agent itself**, while the log where users want to see traceIDs are within their own application logs. You should move these code to the **sw_logging** module under ``skywalking/log``.

CodePrometheus:
Copy that, thanks for the reply, this is my negligence, please give me some time, I will improve this PR as soon as possible.

Superskyyy:
> Copy that, thanks for the reply, this is my negligence, please give me some time, I will improve this PR as soon as possible.

No problem! I will be waiting patiently, take your time. Ping me for any help you need.

wu-sheng:
@Superskyyy As this is a terminal output, we don't have CI to verify this. Could you review and test locally?

wu-sheng:
@CodePrometheus Please fix CI.

CodePrometheus:
The latest CI exception does not seem to be caused by this PR change.
[case1](https://github.com/apache/skywalking-python/actions/runs/6678224210/job/18152337052?pr=323), [case2](https://github.com/apache/skywalking-python/actions/runs/6678224210/job/18152367001?pr=323).

wu-sheng:
@Superskyyy Please recheck, and fix the main branch if needed.

Superskyyy:
Yes, I have a fix coming, it is due to a upstream dependency bug in flask, I will hold the merge of this pr untill that one is fixed first.

Superskyyy:
Code LGTM, please regenerate the configuration documentation by running the checklist command below. I will test it locally tmr.

- [ ] I have rebuilt the `Configuration.md` documentation by running `make doc-gen`

skywalking/log/sw_logging.py
```diff
@@ -75,6 +77,9 @@ def build_log_tags() -> LogTags:
 
         context = get_context()
 
+        if config.agent_log_print_tid:
+            record.tid = str(context.segment.related_traces[0])
```
kezhenxu94, line 81: We add the `tid` info to the `record` only when `agent_log_print_tid == true`, but what if I set `SW_AGENT_LOG_REPORTER_LAYOUT='%(tid)s %(asctime)s [%(threadName)s] %(levelname)s %(name)s - %(message)s` but `agent_log_print_tid == false`? The implementation looks unreasonable to me, we have two ways to modify the log layout which might be conflict with each other.

CodePrometheus: @kezhenxu94 thanks for your comment, my opinion is:
1) If and only if agent_log_print_tid=true, the agent will enable the ability to print traceId.
2) When the user-defined layout has tid, but agent_log_print_tid=false, printing traceId is turned off
3) Or emphasize in the document that when agent_log_print_tid=true, the layout will change accordingly.

Superskyyy: How about we make it more consistent with the Java agent, that is simply introducing the tid mechnism through the custom layout?

kezhenxu94: > How about we make it more consistent with the Java agent, that is simply introducing the tid mechnism through the custom layout?

This makes sense to me

Superskyyy (changes requested):

Superskyyy:
Since tid is injected into the logging record, user could also utilize this to print out the tid in whatever channel they want other than only the reported logs to oap, this can be done using any formatter they use in their own application logic, this should be pointed out in the documentation. @CodePrometheus

Superskyyy (approved):
LGTM

PR #325, opened 2023-10-25
remove the restrict of greenlet

jaychoww:


wu-sheng:
Why do you remove this?

jaychoww:
it seems the gevent 22.10.2

> Why do you remove this?

it seems gevent 22.10.2 fixed the issue about compatibility with greenlet, so I'm trying to do some tests that if the limit can be taken off

jaychoww:
it's

PR #336, opened 2024-04-09
Fix unexpected 'decode' AttributeError when MySQLdb module is mapped by PyMySQL

ZEALi:
In some use cases, `pymysql.install_as_MySQLdb()` will be used to initialize a 'fake' MySQLdb module, which is actually mapped to pymysql module.

Unfortunately, sw_mysqlclient plugin changes the data type of `connection.db` from `bytes` into `str`,  which is regard as `bytes` in sw_pymysql plugin. That will cause the `AttributeError: 'str' object has no attribute 'decode'` exception in sw_pymysql.py line 45.

So if the `MySQLdb` module  is only a pointer to `pymysql`, we should just raise ImportError instead of installation for sw_mysqlclient plugin.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
   - Fix unexpected 'No active span' IllegalStateError (#311)
   - **Tentative**: Set upper bound <=5.9.5 for psutil package due to test failure. (#326)
   - Remove `DeprecationWarning` from `pkg_resources` by replace it with `importlib_metadata` (#329)
+  - Fix unexpected 'decode' AttributeError when MySQLdb module is mapped by PyMySQL (#336)
 
 ### 1.0.1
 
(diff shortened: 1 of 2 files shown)


wu-sheng:
Please fix CI.

wu-sheng:
And you should follow PR template to `Update the CHANGELOG.md.`

wu-sheng:
@kezhenxu94 Please take a look. We also have some CIs fail, but should be irrelevant. We may need some fixes first.

Superskyyy:
Ok I will take a look thanks

Superskyyy:
I will temporariy disable the uwsgi case, I believe that uwsgi is getting less attraction and support will be dropped sooner or later.

Superskyyy:
LGTM, let's wait for CI.

Superskyyy:
Gonna kick off a release soon this weekend.

wu-sheng (approved):

PR #337, opened 2024-04-11
Disable uwsgi e2e

Superskyyy:
The e2e case for uwsgi is always flakey, disabling due to uwsgi is really outdated in the Python ecosystem and should be removed soon. With nogil coming there's no way that uwsgi is going to be widely used in the future.


wu-sheng (approved):

wu-sheng:
If you feel that eco is not useful, please remove them.
If tests can't be maintained, I don't think we should declare we support it.

PR #352, opened 2024-08-10
fix error: in python3.12, AttributeError: 'FileFinder' object has no …

airbot-zn:
fix error: in python3.12, AttributeError: 'FileFinder' object has no attribute 'find_module'


wu-sheng:
Tests fail. Please fix.

wu-sheng:
I am.not sure why you keep changing so many times. Please verify all things locally rather han costing upstream resources and time.

airbot-zn:
on my fork check  test,  I  don't know why it submitted .

PR #371, opened 2025-05-07
chore: cover py3.12 and py3.13 in tests and meta info

shenxiangzhuang:

### DIFF
--- a/.github/workflows/CI.yaml
+++ b/.github/workflows/CI.yaml
@@ -138,7 +138,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix: # may support pypy in the future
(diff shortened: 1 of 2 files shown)


shenxiangzhuang:
I'm not sure about this, but the project file with fellowing config, which means that this package already support python version >=3.8,<=3.13:

https://github.com/apache/skywalking-python/blob/1f091ee33f51fff91dd53e2bb65a15d932a12f89/pyproject.toml#L72

IMHO, these versions should be included in the tests. Is there any history reason that make them can not be included?

wu-sheng:
I can see. It doesn't work. I don't know the agent details, you may need to debug and recheck.

PR #388, opened 2026-04-12
perf(demo): set a timeout on flask consumer fork HTTP calls

basheer-cloud:
This fell out while I was looking at demo/flask_consumer_fork.py. Set a timeout on flask consumer fork HTTP calls. Network calls without a timeout can hang a worker indefinitely. I kept the patch small and re-ran syntax checks after applying it.


wu-sheng:
Could you resolve the conflicts?

kezhenxu94 (approved):
Fix the CI and it LGTM

wu-sheng (approved):

PR #404, opened 2026-06-25
feat(odoo): add Odoo HTTP instrumentation and fail-open bootstrap

0x00kh:
## Summary
- Add Odoo plugin that instruments Root.dispatch to create HTTP entry spans.
- Provide fail-open bootstrap in skywalking.bootstrap.odoo.maybe_start() that starts the agent only when both required env vars are set.
- Add usage docs and unit tests for bootstrap and plugin logic.

## Test Plan
- python -m unittest tests.unit.test_version_check tests.unit.test_odoo_bootstrap tests.unit.test_odoo_plugin -v
- python -m py_compile skywalking/bootstrap/odoo.py skywalking/plugins/sw_odoo.py tests/unit/test_odoo_bootstrap.py tests/unit/test_odoo_plugin.py
- git diff --check HEAD~1..HEAD


wu-sheng:
Please fix the ci and docs.

